PR #1, opened 2021-05-04
base classes, dependencies, and basic tokio runtime first impl

clayrab:
Let's try a code review on this


Cargo.toml
```diff
@@ -1,9 +1,22 @@
 [package]
 name = "saito-rust"
 version = "0.1.0"
-authors = ["Clayton Rabenda <clay@saito.tech>"]
+authors = ["Clayton Rabenda <clay@saito.tech>", "Stephe Peterkins <stephen@saito.tech>"]
```
clayrab, line 4: LOL, well, I didn't think I'd have to review this line but you mis-spelled your own name :P

src/slip.rs
```diff
@@ -1,25 +1,95 @@
...
+    }
+
+    pub fn return_add(&self) -> PublicKey {
+        return self.body.add;
+    }
+
+    // implication is slip type is immutable
+    pub fn return_signature_source(&self) -> Vec<u8> {
+       return bincode::serialize(&self.body).unwrap();
+    }
```
clayrab, line 87: I don't understand this. Perhaps just needs a better comment...

src/transaction.rs
```diff
@@ -1,27 +1,93 @@
...
-// getType()
-// addSlip(Slip)
-// getSlips -> [Slip]
-// pushToPath(address String)
-// createRebroadcastTxFrom() -> Transaction
-// Notable missing things: from, to, timestamp
 
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
+pub struct TransactionBody {
+    id:   u32,
```
clayrab, line 19: Is this a hash?

src/transaction.rs
```diff
@@ -1,27 +1,93 @@
...
+    ts:   u64,
+    pub   to: Vec<Slip>,
+    pub   from: Vec<Slip>,
+    sig:  Signature,
+    ver:  f32,
```
clayrab, line 24: What's ver?

src/transaction.rs
```diff
@@ -1,27 +1,93 @@
...
+    pub typ:  TransactionBroadcastType,
+    path: Vec<Hop>,
```
clayrab, line 26: I honestly have no idea about path and where it should go. This is something I need to understand...

src/transaction.rs
```diff
@@ -1,27 +1,93 @@
...
+    pub msg:  Vec<u8>,
+    ps:   u8,
```
clayrab, line 28: public school? :P

src/block.rs
```diff
@@ -1,34 +1,118 @@
...
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Debug, Clone)]
+pub struct BlockBody {
+    pub id:          u32,
+    pub ts:          u64,
+    pub prevbsh:     [u8; 32],
+    // pub creator:     PublicKey,
+    // pub txs:         Vec<Transaction>,
+    pub bf:	         BurnFee,
```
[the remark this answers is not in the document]
trevelyan: Just as a flag, I'd highly recommend we keep the nomenclature for objects and variables the same between RS and JS and that it is changes from the existing nomenclature that need to be justified. Reason is so that development effort is easier to understand / switch between the two:

bf / ts / blk / etc.

The shorter terms have generally been chosen because we don't want big names for the JSON version (devs should be able to inspect the JSON versions and see what the variables are -- if we start having different names for RS / JS / JSON then things get very, very messy. So lets please flag and discuss changes as a group.

src/block.rs
```diff
@@ -1,34 +1,118 @@
...
+
+use crate::burnfee::BurnFee;
+
+#[derive(Serialize, Deserialize, PartialEq, Debug, Clone)]
 pub struct Block {
-    
+    pub body:     BlockBody,
+    pub is_valid: u8,
+    mintid:   u32,
+    maxtid:   u32,
```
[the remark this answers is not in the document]
trevelyan: I think we use maxtid when creating blocks. The block creator basically needs to know what tid they should use when they start putting things into blocks. it is probably easier to keep it than keep the last TX around and check the ID that way.

Right now mintxid is used in the JS version by the explorer (?). It was added so that the core blockchain.js code could find out what block a particular transaction would be in. And while it might be useful for the engine to be able to scroll through blocks and know where a tx is located, but i don't think we need it for consensus and we can probably drop it in rust if we want.

src/block.rs
```diff
@@ -1,34 +1,118 @@
...
-#[derive(Debug)]
+use serde::{Serialize, Deserialize};
+
+use crate::burnfee::BurnFee;
+
+#[derive(Serialize, Deserialize, PartialEq, Debug, Clone)]
 pub struct Block {
-    
+    pub body:     BlockBody,
+    pub is_valid: u8,
```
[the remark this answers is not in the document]
trevelyan: Just for context, because I have no idea if we want to keep this --- JS uses is_valid as a cheap form of error-catching. Not sure if we still need it in RUST.

The case I'd be concerned about is -- if we create a block from transactions in our mempool and it is invalid because one of the TXS is invalid, do we want to tank all of the transactions in our mempool just because the block is bad?

is_valid is JS-poor-man's way of saying, "just mark the bad ones and we can clean-up later".

Not sure how parallelization works in Rust. It might be useful to keep depend on what happens there, as tagging the block as invalid and then doing a sanity-check at the end when all of the children have finished processing might be cleaner than noping out halfway through. Not sure.

That was the point in JS anyway.

Cargo.toml
BearGuy: LMAO fail

Cargo.toml
```diff
@@ -1,9 +1,22 @@
...
+digest = '0.8'
+rand = "0.4"
+serde = { version = "1.0", features = ["derive"] }
+serde_bytes = "0.11"
+secp256k1 = { version = '0.12.2', features = ["rand", "serde"] }
+base58 = '0.1.0'
+merkle = '1.10.0'
+bincode = "1.1.4"
+ring = "^0.13.0"
+data-encoding = '2.1.2'
```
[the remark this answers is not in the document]
BearGuy: Tokio is up to date, but the rest should be checked. I kept them as-is because I had some versioning trouble the first time around, and I know that this configuration works

src/block.rs
```diff
@@ -1,34 +1,118 @@
...
+    mintid:   u32,
+    maxtid:   u32,
+    bsh:      [u8; 32],
```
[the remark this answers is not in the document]
BearGuy: Yep, the only thing we want to check is how `bincode` handles serialization of longer attribute names vs shorter ones. If there's no difference, longer names are the way to go

Cargo.toml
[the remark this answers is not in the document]
escapedcat: Some with `"` some with `'` some with `^` some with fixed version?  
- Does prettier work with toml files (to keep one code style?)
- Is cargo working with a lock file (in that case `^` makes sense in general, no?)?
- Should we have a look at deps-robots like https://dependabot.com/rust/ ?

src/block.rs
```diff
@@ -1,33 +1,314 @@
...
+impl BlockBody {
+    pub fn new(block_creator: PublicKey, prevbsh: [u8;32]) -> BlockBody {
+        return BlockBody {
+    	    id:          0,
+    	    ts:          create_timestamp(),
+    	    prevbsh:     prevbsh,
+    	    merkle:      [0; 32],
+    	    creator:     block_creator,
+    	    txs:         vec![],
+	        bf:          BurnFee::new(0.0, 0),
```
escapedcat, line 50: Automatic code-formatting would be greta for the whole project. Otherwise i.e. people might be confused if this pupose.

src/block.rs
```diff
@@ -1,33 +1,314 @@
...
+            self.body.difficulty,
+            self.body.paysplit,
+            self.body.vote,
+            self.body.treasury,
+            self.body.coinbase,
+            self.body.reclaimed,
+        );
+    }
+
+    // pub fn get_transactions(&mut self, tx: Transaction) -> Vec<Transaction> {
```
escapedcat, line 90: Comment why this is commented out? Otherwise this will stay forever and people forget about it

src/block.rs
```diff
@@ -1,33 +1,314 @@
...
+                .collect();
+
+            tx.set_from_slips(from_slips);
+
+            // calculate cumulative fees
+            cumulative_fees = tx.calculate_cumulative_fees(cumulative_fees);
+        }
+
+        self.mintid = self.maxtid + 1;
+        self.maxtid = self.maxtid + tx_length  as u32;
```
escapedcat, line 148: Automatic code-formatting would be nice

src/block.rs
```diff
@@ -1,33 +1,314 @@
...
+    }
+
+    pub fn set_coinbase(&mut self, coinbase: u64) {
+        self.body.coinbase = coinbase;
+    }
+
+    pub fn set_reclaimed(&mut self, reclaimed: u64) {
+        self.body.reclaimed = reclaimed;
+    }
+    // fn getTransactionByHash(hash) -> Transaction
```
escapedcat, line 247: See above. Comment why this is commented out.

clayrab:
Closed

PR #7, opened 2021-05-09
Keypair

clayrab:
A basic keypair struct/class.


BearGuy (changes requested):
I would like to propose that PRs be warning free before merge, but overall I like this implementation

Cargo.toml
```diff
@@ -7,3 +7,9 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+secp256k1 = { version = "0.20.2", features = ["rand"] }
```
escapedcat, line 10: Would be nice top use `^` here as a default:  
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

src/lib.rs
```diff
@@ -1,18 +1,26 @@
...
-// pub mod keypair;
-// pub mod crypto;
-// pub mod peer_policy;
-// pub mod concrete_peer_policy;
-// pub mod consensus_policy;
-// pub mod concrete_consensus_policy;
-// pub mod fork_policy;
-// pub mod concrete_fork_policy;
\ No newline at end of file
+/*!
```
escapedcat, line 1: Whats' this doing here?

clayrab: This makes cargo doc pickup this comment and stick it on the main page of the docs. I actually snuck the docs onto production already if you wanna take a look: https://saito.io/docs/saito_rust/

Cargo.toml
clayrab: yes

src/keypair.rs
```diff
@@ -0,0 +1,183 @@
...
+            hashvalue[30],
+            hashvalue[31],
+        ];
+    }
+    /// Hash and sign a string
+    pub fn sign_string_message(&self, message_string: &str) -> String {
+        let message_bytes = Keypair::make_message_from_string(message_string);
+        let bytes = self.sign_message(&message_bytes);
+        let mut string_out = String::new();
+        write!(&mut string_out, "{:?}", bytes);
```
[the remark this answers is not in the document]
clayrab: We changed the method to return a Result... but maybe it would be okay to just expect()... I think I prefer to return a Result but that may be getting too academic, if write! cannot convert bytes to a string perhaps there are bigger problems than behavior or saito-rust...

BearGuy (approved):

PR #10, opened 2021-05-10
Get rid of panics

batzor:
Removed `expect()` where:
- error is not critical
- when error message does not have any meaning


BearGuy:
LGTM

escapedcat:
Awesome, thanks @batzor !

PR #17, opened 2021-05-11
Slip and Transaction

BearGuy:
Introducing the `Slip` and `Transaction` structs. There are additional supporting structs such as `Hop` and enumerated types associated with both structs which are also included in this PR.

I ended up making some functions in `Keypair` public to support usage of `secp256k1::Signature` in `Transaction`, as the current public functions exposed in `Keypair` only allow usage with `String`s and not bytes and signatures directly.


escapedcat:
Checks are failing

src/slip.rs
```diff
@@ -0,0 +1,200 @@
+use secp256k1::{PublicKey};
+
+/// A UTXO slip containing record of owernship of funds on the network
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub struct Slip {
+    /// Contains concrete data which is not relative to state of the chain
+    body: SlipBody,
+    /// Flag determining if the `Slip` is part of the longest chain
+    lc: bool,
```
clayrab, line 9: I'd like to add this when it becomes relevant, i.e. with mempool because I'm not convinced we need it until I see the algorithm that will use it.

src/slip.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// `SlipSpentStatus` indicating the state of the slip in the network
+    pub spent_status: SlipSpentStatus,
```
clayrab, line 11: Same for this, we don't need it until mempool

src/slip.rs
```diff
@@ -0,0 +1,200 @@
...
+}
+
+/// An object that holds concrete data not subjective to state of chain
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub struct SlipBody {
+    /// An `Sectp256K::PublicKey` determining who owns the `Slip`
+    add: PublicKey,
```
clayrab, line 18: Would like to rename this to "to" but i guess we can keep it if this is the same in saito-lite...

src/slip.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// Amount of Saito
+    amt: u64,
+    /// `Block` id
+    bid: u64,
+    /// `Transaction` id
+    tid: u64,
+    /// `Slip` id
+    sid: u64,
+    /// The `Block` hash the slip originated from
+    bsh: [u8; 32],
```
clayrab, line 30: All of these can be computed from the location of the slip in the block encoding. I'm not convinced they even will be needed in the slip metadata until I see mempool algos. I strongly believe we should not include them in the slip body, they will waste bytes on disk and on the wire and make things slower. They can also cause potential bugs because they are duplicated state.

src/slip.rs
```diff
@@ -0,0 +1,200 @@
...
+impl Slip {
+    /// Create new `Slip` with default type `SlipBroadcastType::Normal`
+    ///
+    /// * `add` - `Publickey` address to assign ownership
+    /// * `typ` - `SlipBroadcastType` of `Slip`
+    /// * `amt` - `u64` amount of Saito coantined in the `Slip`
+    pub fn new(add: PublicKey, typ: SlipBroadcastType, amt: u64) -> Slip {
+        return Slip {
+            body: SlipBody {
+                add,
```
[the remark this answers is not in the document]
clayrab: This is an issue we've already resolved, although I agree with you escapedcat. There is a legitimate reason in saito-lite to keep these small because we are using json-encoded blocks and there is an argument that we should keep things looking similar to saito-lite. I personally don't agree with this reasoning but find it acceptable so long as we add comments giving them more expressive names like has been done above.

src/slip.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /// Returns address in `Slip`
+    pub fn get_address(&self) -> PublicKey {
+        return self.body.add;
+    }
+
+    /// Returns`Slip` type from the enumerated set of `SlipBroadcastType`
+    pub fn get_type(&self) -> SlipBroadcastType {
+        return self.body.typ.clone();
```
clayrab, line 81: Does this need to be clone()ed?

src/time.rs
```diff
@@ -0,0 +1,8 @@
+use std::time::{SystemTime, UNIX_EPOCH};
+
+pub fn create_timestamp() -> u64 {
+    let start = SystemTime::now();
+    let since_the_epoch = start.duration_since(UNIX_EPOCH)
+        .expect("Time went backwards");
+    return since_the_epoch.as_millis() as u64;
+}
```
clayrab, line 8: I'd  rather see this as a one-liner and return a Result: 

`pub fn create_timestamp() -> Result<u64> {
SystemTime::now().duration_since(UNIX_EPOCH)?.since_the_epoch.as_millis()
}
`

src/transaction.rs
```diff
@@ -0,0 +1,220 @@
+use secp256k1::{PublicKey, Signature};
+use crate::{
+    slip::{Slip},
+    time::{create_timestamp}
+};
+
+/// A single record used in the history of transactions being routed around the network
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub struct Hop {
+    pub from: PublicKey,
```
clayrab, line 10: I don't think Hop's need a from field to be serialized. This is redudant.

src/transaction.rs
```diff
@@ -0,0 +1,220 @@
...
+/// contains additional information as an optinal message field to transfer data around the network
+#[derive(Debug, PartialEq)]
+pub struct Transaction {
+    body: TransactionBody
+}
+
+/// Core data to be serialized/deserialized of `Transaction`
+#[derive(Debug, PartialEq)]
+pub struct TransactionBody {
+    id:       u64,
```
clayrab, line 35: id is not needed. The order of tx in the block encoding can represent the id.

src/transaction.rs
```diff
@@ -0,0 +1,220 @@
...
+    ts:       u64,
+    pub to:   Vec<Slip>,
```
clayrab, line 37: My intuition is that slips have to fields, but transactions don't need them.

src/transaction.rs
```diff
@@ -0,0 +1,220 @@
...
+    pub msg:  Vec<u8>,
+}
+/// Enumerated types of `Transaction`s to be handlded by consensus
+#[derive(Debug, PartialEq, Clone)]
+pub enum TransactionBroadcastType {
+    Normal,
+    GoldenTicket,
+    Fee,
+    Rebroadcast,
+    VIP,
```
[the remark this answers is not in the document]
clayrab: yeah, a very important transaction. we seed the chain with these.

I'd like to see some comments for each of these explaining what purpose they each serve. I'm personally not even 100% sure about GoldenTicket vs GoldenChunk.

src/slip.rs
```diff
@@ -0,0 +1,200 @@
+use secp256k1::{PublicKey};
+
+/// A UTXO slip containing record of owernship of funds on the network
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub struct Slip {
```
clayrab, line 5: My understanding of UTXO is like this: For the sake of performance, an input/output(what we are calling "slip") must be spent entirely within a transaction.

With this understanding, a slip should have a field called ID or maybe input(e.g. blockid+slip index) which can identify the unspent output that's being spent. it should also have a list(vec or []) of "outputs" which are each address+amount. The work of finding which outputs are appropriate for a given transaction the user wants to do is put onto the wallet, the UTXOSet validation for the node is then just to check for the presence of the outputs in the UTXOSet by ID.

In total, the structures should look like this: A transaction has one sig and multiple slips, a slip can have multiple outputs.

src/transaction.rs
```diff
@@ -0,0 +1,220 @@
...
+    pub to:   Vec<Slip>,
+    pub from: Vec<Slip>,
+    sig:      Signature,
+    pub typ:  TransactionBroadcastType,
+    path:     Vec<Hop>,
+    pub msg:  Vec<u8>,
+}
+/// Enumerated types of `Transaction`s to be handlded by consensus
+#[derive(Debug, PartialEq, Clone)]
+pub enum TransactionBroadcastType {
```
clayrab, line 46: I'm confused why these types are included in both slip and tx. I think these are only needed at the tx level or they should be split into two. e.g. Tx are Normal, GoldenTicket, GoldenChunk Rebroadcast and Slips are Normal or Fee.

Also note that BTC simply doens't include the fee and instead just computes it as the difference between outputs vs input and this saves some space. In this case we could just have Type on tx and no type for slips. I'm probably missing somethign around how fees are paid out, so perhaps that's what "Fee" is...? perhaps we could call it goldenFee or something if that's the case.

src/slip.rs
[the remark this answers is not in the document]
batzor: I think it will be more troublesome to rename those ambiguous down the road so its better to rename when its still early. Adding comments is not going to make it more understandable when its used in modules.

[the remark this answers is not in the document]
BearGuy: I agree with that. If it doesn't become an issue increasing the size when encoding slips, having more specific names is clearly superior

src/transaction.rs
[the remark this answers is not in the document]
BearGuy: We also don't need to list all of them out right this moment. We can introduce them as we build out other algorithms

src/slip.rs
BearGuy: I would prefer to keep it as "add" instead of "to". As a convention, I would prefer to use "to" when there's a "from" field, single fields can be given "add" or "address"

src/slip.rs
BearGuy: Great point. Will move outside of the slip body as we still want to keep it around for checks with `UTXOSet`

src/transaction.rs
BearGuy: This ID isn't in reference to the id in the block but instead a globally unique transaction id. We want to record this as we can't discern it just from the order of just a single block

src/transaction.rs
BearGuy: This would be great for @trevelyan to weigh in on, but I agree it would be nice to simply this by just having either or.

src/slip.rs
BearGuy: I believe so, let me see if I can implement the `Copy` trait on it instead

src/time.rs
BearGuy: This won't compile, but we can simplify the function

src/transaction.rs
BearGuy: We want to keep `Slip`s as simple as possible, and allow `Transaction`s to do the account balancing between the `to` fields and the `from` fields

src/slip.rs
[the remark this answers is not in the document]
batzor: In `bincode`, attribute naming does not affect the encoded size so I think we can go with more specific names.

src/slip.rs
```diff
@@ -1,200 +1,145 @@
...
+            block_hash: [0; 32],
+        };
     }
 
-    /// Returns address in `Slip`
-    pub fn get_address(&self) -> PublicKey {
-        return self.body.add;
+    /// Returns addressress in `Slip`
+    pub fn get_addressress(&self) -> PublicKey {
+        return self.body.address;
```
batzor, line 57: seems like a typo.

BearGuy: yep *facepalm*, fixing now

src/slip.rs
clayrab: Spoke with Stephen and resolved my concerns here. I had a misunderstanding about the to/from fields on transaction.

src/slip.rs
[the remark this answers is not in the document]
clayrab: I agree with batzor on this.

src/transaction.rs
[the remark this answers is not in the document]
clayrab: I'm good with anything here. My concern is really just whether these types go with tx or slip, but i'm pretty sure it should be both.

src/slip.rs
clayrab: Won't we want to call this "spent" rather than "lc" also? let's just tackle this later, there's enough to talk about with just the slip/transaction encoding.

src/slip.rs
clayrab: Spoke to Stephen about this. I was confused because i didn't realize slips are in vectors called from and to at the transaction level, so 'add' here makes sense to me now.

src/slip.rs
clayrab: Thanks

src/slip.rs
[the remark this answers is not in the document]
clayrab: Stephen is working on another branch where we can test bincode vs protobufs vs whatever else and then have some more objective conversation about this.

src/slip.rs
clayrab: Just so it's clear, this is not socratic method, i actually have no idea if it should be clone()ed or not.

src/time.rs
clayrab: just a nitpick, it's fine. I'm actually surprised a function like this does exist in std.

src/transaction.rs
clayrab: Okay, but i we just use {blockid}-{txindex} then that can be used as a tx id which doesn't need to be stored in the tx, right?

src/transaction.rs
clayrab: I had a misunderstanding. Discussed with Stephen. This LGTM.

src/transaction.rs
BearGuy: Good point. Will remove

PR #18, opened 2021-05-11
Burnfee

BearGuy:


src/burnfee.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub current: u64,
+}
+
+impl BurnFee {
+    /// Returns the BurnFee used to calculate the work needed to produce a block
+    ///
+    /// * `start` - y-value at x = 0
+    /// * `current` - y-value at x = 0 for next bloc
+    pub fn new(start: f64, current: u64) -> Self {
+        return Self {
```
escapedcat, line 15: Heard the in the rust community it's more common to use the short return:

```
Self {
  start,
  current
}
```

This can be applied to other `return`s as well.

src/burnfee.rs
```diff
@@ -0,0 +1,106 @@
...
+    ///
+    /// * `prevts` - timestamp of previous block
+    /// * `ts`     - candidate timestamp
+    /// * `start`  - burn fee value (y-axis) for curve determination ("start")
+    pub fn return_work_needed(prevts: u64, ts: u64, start: f64) -> u64 {
+	    let mut elapsed_time = ts - prevts;
+        if elapsed_time == 0 { elapsed_time = 1; }
+        if elapsed_time >= (2 * HEARTBEAT) { return 0; }
+
+        let elapsed_time_float     = elapsed_time as f64;
```
escapedcat, line 34: Is this usage of `as` common in rust?  
Asking because in TS this would be a hint that typing might have issues, because your manually telling it it's `f64`.

batzor: yes it is common and I don't see any problem here.

clayrab:
I would remove "current" from the struct and rename start to "last_block_burnfee".

I think adjust_work_needed isn't needed, return_work_needed should be called each time the current burnfee is needed anyway.

burn_fee_adjustment should be called burn_fee_adjust_forward and have an inverse function called burn_fee_adjustment_rolloback, similar to mempool stuff.

The reasoning behind all this is that burnfee should not be included in blockbody, since it is computable from the timestamps. If nodes want to compute and cache burnfees outside of the blockchain, that's fine, but it's not needed in version 1 and it shouldn't be part of the blockchain itself. For example if you look at bitcoin, i don't think you'll see difficulty in the block encoding.

BearGuy:
> I would remove "current" from the struct and rename start to "last_block_burnfee".
> 
> I think adjust_work_needed isn't needed, return_work_needed should be called each time the current burnfee is needed anyway.
> 
> burn_fee_adjustment should be called burn_fee_adjust_forward and have an inverse function called burn_fee_adjustment_rolloback, similar to mempool stuff.
> 
> The reasoning behind all this is that burnfee should not be included in blockbody, since it is computable from the timestamps. If nodes want to compute and cache burnfees outside of the blockchain, that's fine, but it's not needed in version 1 and it shouldn't be part of the blockchain itself. For example if you look at bitcoin, i don't think you'll see difficulty in the block encoding.

`start` doesn't function as the variable `last_block_burnfee` descrubes, it's the starting point for the burnfee algorithm (the y-axis value specifically in the equation). We could call it `starting_burnfee` to be more specific.

We can still calculate the burnfee that was paid based on the timestamps, so I agree we can drop `current`.

We could keep the `BurnFee` object start under `Mempool` instead of on each block. It'll contain the `starting_burnfee` value, and it can mutate that value as `Mempool` sees fit (whether going forward or rolling back).

src/burnfee.rs
clayrab: There's a bigger issue here which is that I think this should be an int, not a float....

PR #19, opened 2021-05-13
Clean up so merge to release pass checks

batzor:
- Rebased and fixed the tests to make merging smoother

ADDITIONAL NOTE: From [Rust naming convention](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter), we should use ie `address()` instead of `get_address()`. Also, it is generally better practice for performance to return references for the complex structures (ie `Vec`) instead of cloning them inside the getter and clone it after the getter function return if cloning is necessary.


BearGuy (approved):
LGTM 👍

PR #21, opened 2021-05-13
Fixes to docs

BearGuy:


README.md
```diff
@@ -43,13 +39,13 @@ Format code according to the [Rust style Guide](https://github.com/rust-dev-tool
 
 GH Actions are located here: [.github/workflows](.github/workflows)
 
-- cargo docs  
+- cargo docs
```
escapedcat, line 42: @BearGuy we need these two spaces `  `. without those the list line-break will not work

escapedcat: current state:  
image

escapedcat: do you use a markdown-formatter in your ide?

PR #22, opened 2021-05-14
Basic Block Struct

BearGuy:
First implementation of our `Block` struct, complete with `BlockBody`. We're also introducing the `crypto` class in here as well, which we'll be using for our `hash` function in a few places in the future.

`set_transactions` is possibly a bit more complex than it needs to be at this stage, but given we already have `Slip` and `Transaction` implemented it made sense to me to fill it out a little bit more.


src/block.rs
```diff
@@ -0,0 +1,314 @@
...
+    pub previous_block_hash: [u8; 32],
+    /// `Publickey` of the block creator
+    pub creator: PublicKey,
+    /// List of transactions in the block
+    pub txs: Vec<Transaction>,
+    /// `BurnFee` containing the fees paid to produce the block
+    pub burnfee: u64,
+    /// Block difficulty required to win the `LotteryGame` in golden ticket generation
+    difficulty: f32,
+    /// Ratio determing the block reward split between nodes and miners
```
clayrab, line 34: I believe this is what we've been calling "advanced saito" and shouldn't be in the first release.

src/block.rs
```diff
@@ -0,0 +1,314 @@
...
+    paysplit: f32,
```
clayrab, line 35: I believe this is what we've been calling "advanced saito" and shouldn't be in the first release.

src/block.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// Vote determining change of difficulty and paysplit
+    vote: i8,
```
clayrab, line 37: I believe this is what we've been calling "advanced saito" and shouldn't be in the first release.

src/block.rs
```diff
@@ -0,0 +1,314 @@
...
+    bsh: [u8; 32],
+}
+/// This `BlockBody` holds data to be serialized along with
+/// `Transaction`s
+#[derive(PartialEq, Debug, Clone)]
+pub struct BlockBody {
+    /// Block id
+    pub id: u64,
+    /// Block timestamp
+    pub ts: u64,
```
escapedcat, line 23: can this be named `timestamp`?

src/block.rs
```diff
@@ -0,0 +1,314 @@
...
+
+        return hash(data);
+    }
+
+    /// Converts our blockhash from a byte array into a hex string
+    pub fn block_hash_hex(&self) -> String {
+        hex::encode(&self.block_hash())
+    }
+
+    /// Swaps a list of transactions already in memory into blocks transactions
```
escapedcat, line 166: "...into blocks transactions"? is that a typo?

src/block.rs
```diff
@@ -0,0 +1,314 @@
...
+    ///
+    /// The transactions are given the proper ids in reference to other transactions
+    /// already on chain, as well as setting the proper metadata for each slip in
+    /// each transaction
+    pub fn set_transactions(&mut self, transactions: &mut Vec<Transaction>) {
+        // Memory swap of transactions so we don't have to copy large amounts of data twice
+        std::mem::swap(&mut self.body.txs, transactions);
+
+        // let maxtid = self.maxtid;
```
escapedcat, line 175: is this needed?

src/block.rs
```diff
@@ -0,0 +1,314 @@
...
+        self.body.treasury = treasury;
+    }
+
+    /// Sets the `Block` coinbase
+    pub fn set_coinbase(&mut self, coinbase: u64) {
+        self.body.coinbase = coinbase;
+    }
+}
+
+// Module std::stream
```
escapedcat, line 257: is this needed?

src/block.rs
BearGuy: yep!

src/block.rs
BearGuy: yeah I'll clean up the wording on some of the documentation

PR #24, opened 2021-05-14
Optimize set_transactions().

batzor:
- Changed few functions to return reference of the object without cloning as I believe most use cases will be read-only. Cloning can be done if necessary after the function call.
- Optimize and simplify `set_transactions()` function to use no memcpy whereas [`std::mem::swap` uses temporary buffer to do the swapping](https://doc.rust-lang.org/src/core/ptr/mod.rs.html#518). It is indeed a useful function if we need the old transactions in `self.body.txs` but we don't. So it's safe to assign the new transactions. It should do the same thing as previous function and the tests pass. But currently, test for `set_transactions()` is too shallow and I don't actually get the logic behind setting `current_sid`.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -97,13 +97,13 @@ impl Block {
     }
 
     /// Returns the previous `Block` hash
-    pub fn previous_block_hash(&self) -> [u8; 32] {
-        self.body.previous_block_hash
+    pub fn previous_block_hash(&self) -> &[u8; 32] {
+        &self.body.previous_block_hash
     }
 
     /// Returns the `Block` creator's `secp256k1::PublicKey`
-    pub fn creator(&self) -> PublicKey {
-        self.body.creator
+    pub fn creator(&self) -> &PublicKey {
+        &self.body.creator
     }
 
     /// Returns the `Block`'s `Transaction`s
@@ -155,7 +155,7 @@ impl Block {
         data.extend(&ts_bytes);
         data.extend(&cr_bytes);
 
-        return hash(data);
+        return hash(&data);
     }
 
     /// Converts our blockhash from a byte array into a hex string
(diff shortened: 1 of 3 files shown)


batzor:
I also renamed some fields and fixed some of the typos. 
`from()` can be easily mistaken for [`From` trait](https://doc.rust-lang.org/std/convert/trait.From.html) so I renamed `from` => `senders` and `to` => `receivers`. It also helps adding clarity to some of the functions such as `add_to_slips` => `append_receiver`.

clayrab:
Can you split this into a few PRs? The optimizations are easy to accept, but the renames will need to get signoff from the rest of the team.

The names of the slip/transaction body fields(ts, sig, etc) were kept small to match saito-lite, which is currently using a json-encoded block, so long names there are a problem. FWIW, I would rather just user the longer names like you've done here but this was already discussed and rejected.

I also think transaction to/from should not be changed to senders/receivers. These are slips(inputs/outputs), not pubkeys, so the names sender/receiver don't make sense to me.

If you want to suggest new names for things, I'd say put them in small PRs. If they are controversial, they're more likely to get accepted if they're smaller and we can address one or two at a time.

batzor:
@clayrab I thought we settled on using descriptive names following PR #17 and there has already been changes made by @BearGuy regarding that. About slips, yeah I am not entirely sure what slip is so it was suggestive edit, I will reverse it for now. But I think `to` and `from` is ambiguous and overlaps with some traits causing confusion so maybe changing it to `inputs` and `outputs` is better.

BearGuy:
I agree with changing `to` and `from` to `inputs` and `outputs`. Way easier to understand and avoids conflict, seems like win win to me.

BearGuy (approved):
LGTM

PR #29, opened 2021-05-17
Tokio runtime and basic consensus runtime implementation

BearGuy:
Creating the foundation for starting to run the Saito server. Perhaps there's a bit we can trim with regards to the `channels`, but wanted to introduce it now just to lay the pipes for graceful shutdown from the get go

### DIFF
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -1,7 +1,0 @@
-use saito_rust::server;
-use tokio::signal;
-
-#[tokio::main]
-pub async fn main() -> saito_rust::Result<()> {
-    server::run(signal::ctrl_c()).await
-}
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -1,0 +1,7 @@
+use saito_rust::server;
+use tokio::signal;
+
+#[tokio::main]
+pub async fn main() -> saito_rust::Result<()> {
+    server::run(signal::ctrl_c()).await
+}
--- a/src/server.rs
+++ b/src/server.rs
@@ -1,0 +1,60 @@
+use std::future::Future;
+use tokio::sync::{
+    broadcast,
+    broadcast::{Receiver, Sender},
+    mpsc,
+};
+/// The server state which exposes a run method
+/// initializes pre-connection state
+struct Server {
+    /// Broadcasts a shutdown signal to all active connections.
+    _notify_shutdown: broadcast::Sender<()>,
+    /// Used as part of the graceful shutdown process to wait for client
+    /// connections to complete processing.
+    _shutdown_complete_rx: mpsc::Receiver<()>,
+    _shutdown_complete_tx: mpsc::Sender<()>,
+}
+
+/// Run the Saito server
+pub async fn run(shutdown: impl Future) -> crate::Result<()> {
+    // When the provided `shutdown` future completes, we must send a shutdown
+    // message to all active connections. We use a broadcast channel for this
+    // purpose. The call below ignores the receiver of the broadcast pair, and when
+    // a receiver is needed, the subscribe() method on the sender is used to create
+    // one.
+    let (notify_shutdown, _) = broadcast::channel(1);
+    let (shutdown_complete_tx, shutdown_complete_rx) = mpsc::channel(1);
+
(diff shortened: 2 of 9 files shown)


batzor:
I am not sure why graceful shutdown is necessary. Nodes cannot trust that other nodes will gracefully shutdown so we will  have to remove it from peers with some sort of timeout mechanism anyways. Is there any benefits with implementing the graceful shutdown?

batzor:
By server, you are talking about the node in the Saito P2P blockchain right? In a p2p structure, a single node is both acting as client and server so maybe it's a bit misleading. Or am I missing something here?

src/server.rs
```diff
@@ -0,0 +1,60 @@
...
+    let mut server = Server {
+        _notify_shutdown: notify_shutdown,
+        _shutdown_complete_tx: shutdown_complete_tx,
+        _shutdown_complete_rx: shutdown_complete_rx,
+    };
+
+    tokio::select! {
+        res = server._run() => {
+            if let Err(err) = res {
+                // TODO -- implement logging/tracing
```
escapedcat, line 37: You wanna create an issue or TODO-task (Here: https://github.com/orgs/SaitoTech/projects/5) and link it here in teh comment?

src/server.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// Run the Saito server
+pub async fn run(shutdown: impl Future) -> crate::Result<()> {
+    // When the provided `shutdown` future completes, we must send a shutdown
+    // message to all active connections. We use a broadcast channel for this
+    // purpose. The call below ignores the receiver of the broadcast pair, and when
+    // a receiver is needed, the subscribe() method on the sender is used to create
+    // one.
+    let (notify_shutdown, _) = broadcast::channel(1);
+    let (shutdown_complete_tx, shutdown_complete_rx) = mpsc::channel(1);
```
clayrab, line 26: Can we add an ENUM to hold all these channel numbers??

What is 1?? the shutdown notification channel?

src/server.rs
clayrab: Adding a logger would be a good idea....

src/bin/server.rs
```diff
@@ -0,0 +1,7 @@
+use saito_rust::server;
+use tokio::signal;
+
+#[tokio::main]
+pub async fn main() -> saito_rust::Result<()> {
+    server::run(signal::ctrl_c()).await
+}
```
clayrab, line 7: Should we just put this in main.rs??

I definitely don't like src/bin/, I'd expect to find binaries in there....

Are we going to have more than one entry point? I think you can just replace main.rs with this...

BearGuy:
> I am not sure why graceful shutdown is necessary. Nodes cannot trust that other nodes will gracefully shutdown so we will have to remove it from peers with some sort of timeout mechanism anyways. Is there any benefits with implementing the graceful shutdown?

I still think for our benefit it's useful to have a graceful shutdown. If we were in the middle of bundling a block or some other important operation, we'd want to be handle a shutdown gracefully without putting ourselves into an awkward, untracked state that could cause bugs

BearGuy:
> By server, you are talking about the node in the Saito P2P blockchain right? In a p2p structure, a single node is both acting as client and server so maybe it's a bit misleading. Or am I missing something here?

Hmm, this is a good point. The naming's not quite right for what we want it to do. Basically we want some sort of "consensus" environment to initialize everything and get it all running and configured. I'm thinking we can simply call this environment "consensus"

src/server.rs
BearGuy: Good idea, will implement it

src/bin/server.rs
BearGuy: the `src/bin` setup allows us flexibility in kicking off the system. In the future if we wanted to implement a cli, we could implement a `src/bin/cli.rs` to help facilitate that.

Perhaps this is getting a bit ahead of ourselves however, so for now we'll just settle for a classic `main.rs`

src/server.rs
BearGuy: `channel` has a buffer argument, meaning it can only support one message at a time (which is all we should need when shutting down).

PR #30, opened 2021-05-17
Mempool implementation

BearGuy:


src/mempool.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn work_available(&self) -> u64 {
+        self.work_available
+    }
+
+    /// Attempt to create a new `Block`
+    ///
+    /// * `previous_block` - `Option` of previous `Block`
+    fn _try_bundle(&mut self, previous_block: Option<Block>) -> Option<Block> {
+        if self._can_bundle_block(previous_block) {
+            Some(self._bundle_block(None))
```
batzor, line 47: Shouldn't you pass the previous block instead of `None`?

src/mempool.rs
```diff
@@ -0,0 +1,156 @@
...
+            }
+            None => {
+                block = Block::new(publickey.clone(), [0; 32]);
+            }
+        }
+
+        block.set_transactions(&mut self._transactions);
+
+        return block;
+        // TODO -- calculate difficulty and paysplit changes
```
escapedcat, line 108: See comment regarding TODOs in other PR

src/mempool.rs
BearGuy: oops, yes it should! *facepalm*

PR #31, opened 2021-05-18
Blockchain Implementation

BearGuy:
Just the basics so that we can merge with `Mempool` #30 under some sort of runtime to be determine in #29.

### DIFF
--- a/src/blockchain.rs
+++ b/src/blockchain.rs
@@ -1,11 +1,11 @@
-use crate::block::Block;
-use std::collections::HashMap;
+use crate::block::{Block, BlockHeader};
 
+pub type BlockIndex = (BlockHeader, [u8; 32]);
 /// Indexes of chain attribute
 #[derive(Debug, Clone)]
 pub struct BlockchainIndex {
     /// Vector of blocks
-    blocks: Vec<Block>,
+    blocks: Vec<BlockIndex>,
 }
 
 impl BlockchainIndex {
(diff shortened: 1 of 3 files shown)


src/blockchain.rs
```diff
@@ -0,0 +1,181 @@
+use crate::block::Block;
+use std::collections::HashMap;
+
+/// Indexes of chain attribute
+#[derive(Debug, Clone)]
+pub struct BlockchainIndex {
```
clayrab, line 6: I don't think we can possibly keep this all in memory... If we need a way to access blocks from deep down the chain we'll need to make some sort of lazy-loading cache or something...

clayrab: Maybe it's okay to keep this around for now, btu we need an interface to this structure that's going to allow us to store these on disk...

src/blockchain.rs
```diff
@@ -0,0 +1,181 @@
...
+    last_burnfee: f64,
+}
+
+impl Blockchain {
+    /// Create new `Blockchain`
+    pub fn new() -> Self {
+        Blockchain {
+            index: BlockchainIndex::new(),
+            block_hash_longest_chain_hashmap: HashMap::new(),
+            block_hash_block_id_hashmap: HashMap::new(),
```
clayrab, line 59: I think we should remove these data structures until they are needed.

If you think we need it very soon, can you give me an idea what they're needed for?

src/blockchain.rs
```diff
@@ -0,0 +1,181 @@
...
+
+            longest_chain_position: 0,
+            longest_chain_is_set: false,
```
clayrab, line 62: I'm scared of race conditions and bugs that could be introduced by storing state like this...

I'd rather maybe just see this class have an initialization method that blocks everything and is run first in main()...? But I'm not relaly sure what purpose this bool serves, so perhaps that woudln't work...

src/blockchain.rs
```diff
@@ -0,0 +1,181 @@
...
+/// longest-chain as well as the material that is sitting off
+/// the longest-chain but capable of being switched over.
+#[derive(Debug, Clone)]
+pub struct Blockchain {
+    /// Index of `Block`s
+    index: BlockchainIndex,
+    /// Hash map of all `Block`s received, both on and off the longest chain
+    block_hash_longest_chain_hashmap: HashMap<[u8; 32], u8>,
+    /// Hash map of the ids of `Block` hashes
+    block_hash_block_id_hashmap: HashMap<[u8; 32], u64>,
```
clayrab, line 29: I think we should remove these data structures until they are needed.

If you think we need it very soon, can you give me an idea what they're needed for?

src/blockchain.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    /// Position of the longest chain
+    longest_chain_position: usize,
+    /// Flag indicating if the lognest chain is set
+    longest_chain_is_set: bool,
```
clayrab, line 34: I'm scared of race conditions and bugs that could be introduced by storing state like this...

I'd rather maybe just see this class have an initialization method that blocks everything and is run first in main()...? But I'm not relaly sure what purpose this bool serves, so perhaps that woudln't work...

src/blockchain.rs
clayrab: LGTM

PR #32, opened 2021-05-18
UTXOset Implementation

BearGuy:
The beginning of UTXOset to get all of the pieces we need for the first milestone. 

The last part of this before it's finished is determining how we're converting `Slip` into a byte array (`Vec<u8>`). We can either create a basic placeholder, or we can just hash the value and keep things simple.

If we're hashing for now, then I want to wait to see how @batzor implements the `Hash` trait for `Block` and do something similar for `Slip`.

### DIFF
--- a/src/slip.rs
+++ b/src/slip.rs
@@ -1,8 +1,8 @@
-use crate::crypto::hash;
 use secp256k1::PublicKey;
+use std::hash::Hash;
 
 /// A record of owernship of funds on the network
-#[derive(Debug, PartialEq, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq)]
 pub struct Slip {
     /// Contains concrete data which is not relative to state of the chain
     body: SlipBody,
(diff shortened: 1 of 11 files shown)


batzor:
Can you take a look at `https://doc.rust-lang.org/std/hash/trait.Hash.html`? If all the fields have Hash trait it will derive hash function by just calling hash on every field. Of course we can override it to include only few specific fields.

src/slip.rs
```diff
@@ -87,22 +88,37 @@ impl Slip {
...
+
+    /// Set the `Block` id
     pub fn set_block_id(&mut self, block_id: u64) {
         self.block_id = block_id;
     }
 
-    // Set the `Transaction` id
+    /// Set the `Transaction` id
     pub fn set_tx_id(&mut self, tx_id: u64) {
         self.tx_id = tx_id;
```
escapedcat, line 113: will this stay `tx` or also changed to `transaction`?

BearGuy: this could change, but that refactor is outside of the scope of this PR

PR #33, opened 2021-05-19
Refactor/blocks

batzor:
- Introduce `BlockHeader` to separate block metadata
- Rename `txs` to `transactions` to be consistent with `mempool.rs`
- Fix unreliable hash memoization. Instead of removing it completely, I reset the cached hash value upon update of any field.
- Remove unsafe{} calls in `hash()`. I didn't change the hashing function yet..
- Remove block_hash from Slip as it is hard to update it all the time and I believe it is a bad design.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -1,5 +1,3 @@
-use std::mem::transmute;
-
 use crate::crypto::{hash, PublicKey};
 use crate::time::create_timestamp;
 use crate::transaction::Transaction;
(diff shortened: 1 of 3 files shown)


src/block.rs
```diff
@@ -37,120 +35,139 @@ pub struct BlockBody {
...
 }
 
-impl Block {
-    /// Receives the a publickey and the previous block hash
+impl BlockBody {
+    /// Creates a new `BlockBody`
     ///
-    /// * `block_creator` - `secp256k1::PublicKey` of the block creator
+    /// * `creator` - `secp256k1::PublicKey` of the block creator
     /// * `previous_block_hash` - Previous block hash in bytes
```
BearGuy, line 69: Arguments in documentation need to be removed if they're not function arguments

BearGuy (changes requested):
Overall, LGTM. Just address the small change in documentation then it's good to go

BearGuy (approved):

PR #34, opened 2021-05-21
Feature: mempool block bundling in consensus runtime

BearGuy:
Not complete just yet, still need to add the Blockchain PR #31 to finish this off. The overall structure can be commented on before it's completed.

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -1,9 +1,7 @@
-use std::future::Future;
-use tokio::sync::{
-    broadcast,
-    broadcast::{Receiver, Sender},
-    mpsc,
-};
+use crate::{blockchain::Blockchain, keypair::Keypair, mempool::Mempool, types::SaitoMessage};
+use std::{future::Future, thread::sleep, time::Duration};
+use tokio::sync::{broadcast, mpsc};
+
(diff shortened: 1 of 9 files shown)


src/consensus.rs
```diff
@@ -50,11 +48,26 @@ pub async fn run(shutdown: impl Future) -> crate::Result<()> {
 impl Consensus {
     /// Run consensus
     async fn _run(&mut self) -> crate::Result<()> {
-        let (_tx, mut rx): (Sender<bool>, Receiver<bool>) = broadcast::channel(1);
+        let (saito_message_tx, mut saito_message_rx) = broadcast::channel(32);
```
clayrab, line 51: Would like to just call these sender and receiver rather than tx and rx.

clayrab (commented):
Just merge this. Only things left are a few nitpicks. I got a merge conflict. I'd normally try to resolve it but it's getting late.

src/consensus.rs
BearGuy: "Tx" and "rx" stand for "transmitter" and "receiver" and are the common nomenclature used with tokio channels. I would prefer to stick with convention on this one

PR #35, opened 2021-05-21
Feat/merkle

batzor:
- Merge with `serializers` branch to serialize transactions (I think its fine to have it in the main branch as well)
- Construct merkle tree from `Vec<Transactions>` and memoize it in `Block`. We can change the serializing method by only changing one line of code so I think my choice of `bincode` is not that important.


clayrab:
Hey Batzor,

Thanks for the PR. I don't want to merge this because the serializers branch was only meant to be used as a test space for testing protobufs vs bincode vs custom serialization.

If you could possibly move the merkle root stuff to a separate PR that would be helpful. Otherwise, we'll need to clean up this branch after we settle on a serialization method before we merge it. I don't want to merge a bunch of code that ultimately just needs to be deleted.

PR #48, opened 2021-05-26
Feature: storage struct for async reading and writing blocks

BearGuy:
First implementation of `Storage` which reads and writes blocks asynchronously. We implement basic serialization with `Block`, which we black box behind the `From` and `Into` traits so it's trivial to swap out for a different method.


clayrab:
I don't understand why we would have a namespace "Storage" and keep everything related to storage in that place. I think each thing that needs to be stored should be responsible for it's own storage. i.e. write_block_to_disk and read_block_from_disk should be in block.rs or maybe blockchain.rs.

BearGuy:
> I don't understand why we would have a namespace "Storage" and keep everything related to storage in that place. I think each thing that needs to be stored should be responsible for it's own storage. i.e. write_block_to_disk and read_block_from_disk should be in block.rs or maybe blockchain.rs.

The purpose of having Storage as its own namespace is because we want this functionality to be modularized and reusable in at least two places, we want to design namespaces that achieve specific goals in their functionality, and because we want to uphold the purity of data primitives. 

Pretty much the only objects we want to read/write to disk are just `Block`s. Maybe `Wallet` or `Keypair` at some point in the future, but not currently. So we're mainly talking about `Block`s when we're talking about `Storage`.

`Blockchain` is not the only place where we want to read and write `Block`s to disk. The `Network` namespace will also want this sort of control too. Having one place that controls this functionality is not only more reusable and cleaner, but it stops us from needing to copy any additional storage configuration in two places.

Generally, I am against primitives controlling higher level functionality. I believe that `block.rs` should remain as simple as possible, and solely act as a way to contain block data with some setters and getters. Instead, higher level namespaces should be used to act on `Block` data, which includes both `Storage` for read/write and in the future `Validation` to validate `Block`s.

Even if we only read and write `Block`s in `Blockchain`, it would be better to have a separate namespace that would handle those operations in the same way we implement `UTXOSet` to handle `Slip` validation. This allows us to use each namespace in a way that it achieves a specific goal, and I'd argue overall that achieves a cleaner design. If we were to go in this direction, perhaps we could rename the namespace it to be more specific, such as `BlockStorage` instead of just `Storage` to reflect this specificity.

TLDR: A Storage namespace compartmentalizes this functionality in a way that's more reusable than ad hoc implementations, achieves better design that keeps namespaces focused in their implementation, and upholding the purity of data primitives.

clayrab:
>  we want this functionality to be modularized and reusable in at least two places


If we take the current PR as an example, I don't see how either storage_write_block_to_disk or storage_read_block_to_disk will be reusable.

> we want to design namespaces that achieve specific goals in their functionality

I'd argue that the namespace blockchain is more meaningful than storage, but let's call this a draw. Ultimately if I were freshly coming to the project and I wanted to know how blocks are stored i'd look for something called block

and because we want to uphold the purity of data primitives.


I'd argue that "storage" is 
> Pretty much the only objects we want to read/write to disk are just `Block`s. Maybe `Wallet` or `Keypair` at some point in the future, but not currently. So we're mainly talking about `Block`s when we're talking about `Storage`.
> 
> `Blockchain` is not the only place where we want to read and write `Block`s to disk. The `Network` namespace will also want this sort of control too. Having one place that controls this functionality is not only more reusable and cleaner, but it stops us from needing to copy any additional storage configuration in two places.
> 
> Generally, I am against primitives controlling higher level functionality. I believe that `block.rs` should remain as simple as possible, and solely act as a way to contain block data with some setters and getters. Instead, higher level namespaces should be used to act on `Block` data, which includes both `Storage` for read/write and in the future `Validation` to validate `Block`s.
> 
> Even if we only read and write `Block`s in `Blockchain`, it would be better to have a separate namespace that would handle those operations in the same way we implement `UTXOSet` to handle `Slip` validation. This allows us to use each namespace in a way that it achieves a specific goal, and I'd argue overall that achieves a cleaner design. If we were to go in this direction, perhaps we could rename the namespace it to be more specific, such as `BlockStorage` instead of just `Storage` to reflect this specificity.
> 
> TLDR: A Storage namespace compartmentalizes this functionality in a way that's more reusable than ad hoc implementations, achieves better design that keeps namespaces focused in their implementation, and upholding the purity of data primitives.

clayrab:
>  we want this functionality to be modularized and reusable in at least two places

If we take the current PR as an example, I don't see how either storage_write_block_to_disk or storage_read_block_to_disk will be reusable.

> we want to design namespaces that achieve specific goals in their functionality

I'd argue that the namespace blockchain is more meaningful than storage. Ultimately if I were freshly coming to the project and wanted to know how blocks are stored i'd look for something called "block" not something called "storage". I think this is similar to having a namespace like "helpers", it's too generic and there is no state that the namespace manages, i.e. there is no "separation of responsibilities" because, in my view, "responsibilities" are state.

> and because we want to uphold the purity of data primitives.

I'm not sure what "purity of data primitives" means. The state of block.rs would not be any different, i'm just suggesting that the methods related to storing and reading blocks from disk could be in the block namespace. The runtime structs and serialized blocks would not be affected.

> ...
> Having one place that controls this functionality is not only more reusable and cleaner, but it stops us from needing to copy any additional storage configuration in two places.

I don't see any possiblity of reuse here unless we have a function like "store_u8array_to_location(...)". 

Also, putting all storage functions in once place is not the only way to avoid copying configuration.

> ...

Anyway, all that being said, I guess having a "storage" namespace is fine. I'd just like to see it looking more reusable if that's the goal.

I'd also like to see a Config namespace that handles all configuration of things like "./data/blocks/" which can all be overridden in a standard way for all config setting in the entire process, e.g. via a single config file or command line arguments.

src/storage.rs
```diff
@@ -0,0 +1,82 @@
...
+        };
+
+        filename.push_str(&block.hash_as_hex());
+        filename.push_str(&".sai");
+
+        println!("{:?}", filename);
+
+        let mut buffer = File::create(filename).await?;
+
+        let byte_array: Vec<u8> = block.into();
```
clayrab, line 33: Can we please do something like block.serialize() instead of block.into()?

I'm okay with using serde/bincode for this for now, but we need to wrap it in a function somewhere, imo, even if it just does block.into() for now.

PR #56, opened 2021-05-28
Feature: add block to blockchain first implementation

BearGuy:
This is the first part of adding a block to blockchain, with large parts of the algorithm not implemented. This PR is mostly a proposal for setting data structures needed to implement the algorithm and setting the layout.

Still a work in progress, so this will be a draft until the details are worked through.

### DIFF
--- a/src/blockchain.rs
+++ b/src/blockchain.rs
@@ -1,19 +1,38 @@
 use crate::block::{Block, BlockHeader};
 use crate::utxoset::UTXOSet;
+use std::collections::HashMap;
 
+/// We want to keep information on the block header, the SHA256hash,
+/// and whether or not the block is on the longest chain
 pub type BlockIndex = (BlockHeader, [u8; 32]);
 /// Indexes of chain attribute
 #[derive(Debug, Clone)]
 pub struct BlockchainIndex {
-    /// Vector of blocks
-    blocks: Vec<BlockIndex>,
+    /// Map of Blocks
+    block_map: HashMap<[u8; 32], BlockIndex>,
+    // HashMap of `Block`s that are on the longest chain
+    longest_chain_map: HashMap<[u8; 32], bool>,
+    /// The block has of the latest `Block` on the longest chain
+    last_block_hash: Option<[u8; 32]>,
+    /// The `Block` at the end of the epoch
+    genesis_block_hash: Option<[u8; 32]>,
 }
 
 impl BlockchainIndex {
     /// Creates new `BlockahinIndex`
     pub fn new() -> Self {
-        BlockchainIndex { blocks: vec![] }
+        BlockchainIndex {
+            block_map: HashMap::new(),
+            longest_chain_map: HashMap::new(),
+            last_block_hash: None,
+            genesis_block_hash: None,
+        }
     }
+}
+
+pub struct ChainFork {
+    blocks: Vec<BlockIndex>,
+    _cumulative_burnfee: u64,
 }
 
 /// The structure represents the state of the
@@ -22,7 +41,7 @@ impl BlockchainIndex {
 /// the longest-chain but capable of being switched over.
 #[derive(Debug, Clone)]
 pub struct Blockchain {
-    /// Index of `Block`s
+    /// A set of indices used to track the state of the `Blockchain`
     index: BlockchainIndex,
     /// Hashmap of slips used by the network
     utxoset: UTXOSet,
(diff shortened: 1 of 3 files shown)


src/blockchain.rs
```diff
@@ -39,20 +58,169 @@ impl Blockchain {
...
     }
 
     /// Append `Block` to the index of `Blockchain`
     ///
     /// * `block` - `Block` appended to index
-    pub fn add_block(&mut self, block: Block) {
+    pub fn try_add_block(&mut self, block: Block) -> crate::Result<()> {
+        // TODO -- we want to lock writes to the blockchain while we're adding a block
+        // To do that, we should implement a RwLock around the `Blockchain` index to verify
+        // that there's only one thread that is writing to the chain at once time
```
clayrab, line 80: I don't think anyone should read from the blockchain during this time either because of things that look similar to the "ABA problem".

Put another way, the state of blockchain is inconsistent during add_block and I think even reading from it could be dangerous...

src/blockchain.rs
```diff
@@ -1,28 +1,47 @@
...
-    /// Vector of blocks
-    blocks: Vec<BlockIndex>,
+    /// Map of Blocks
+    block_map: HashMap<[u8; 32], BlockIndex>,
+    // HashMap of `Block`s that are on the longest chain
+    longest_chain_map: HashMap<[u8; 32], bool>,
+    /// The block has of the latest `Block` on the longest chain
+    last_block_hash: Option<[u8; 32]>,
+    /// The `Block` at the end of the epoch
+    genesis_block_hash: Option<[u8; 32]>,
```
clayrab, line 18: I was hoping to refactor block.rs before we do this work because it woudl be easier to do in that order... this is fine...

My intuition is that longest_chain should be a stack of block_hash(the height of the stack is the block_id). last_block_hash is then just a method that does longest_chain_stack.peek(). 

I'm still confused everytime i see the word "genesis" in a Saito context... What does genesis_block_hash represent? Will this get changed every time we do add_block?

We'll need a way to track the bottom of the epoch for rebroadcasting. Is this the purpose of genesis_block_hash, I think we again would need a data structure like block_id -> block_hash. The stack i mentioned above almost works, but it needs to also work like a queue. I.E. we need a to store block hashes and have both Queue(FIFO) and a Stack(LIFO) properties, this is a Double-ended queue. This seems correct to me because it's basically what an epoch looks like.

So, I think we need the block_map as is and everything else can be managed by a Deque:

epoch_block_hashes: Deque;

std seems to have a VecDeque, but we should do a bit more research before we choose which deque to use. https://doc.rust-lang.org/std/collections/struct.VecDeque.html

src/blockchain.rs
```diff
@@ -39,20 +58,169 @@ impl Blockchain {
...
     /// * `block` - `Block` appended to index
-    pub fn add_block(&mut self, block: Block) {
+    pub fn try_add_block(&mut self, block: Block) -> crate::Result<()> {
+        // TODO -- we want to lock writes to the blockchain while we're adding a block
+        // To do that, we should implement a RwLock around the `Blockchain` index to verify
+        // that there's only one thread that is writing to the chain at once time
+
+        if let Some((genesis_block_header, _)) = self.get_genesis_block_index() {
+            // check if our candidate blocks is greater than the genesis block
+            if self.genesis_block_check(&genesis_block_header, &block) {
```
clayrab, line 84: I'd like to do all validation-like things prior to doing add_block, or, at least, I'd like to put these things into a separate function, regardless of how the functions are called.

src/blockchain.rs
```diff
@@ -39,20 +58,169 @@ impl Blockchain {
...
+                // Add the block to the index, but not the longest chain just yet
+                self.add_block_index(&block);
+
+                // If our candidate block previous block_hash is equal to last_block_hash,
+                // then the candidate block is building on the longest chain
+                if &block.header().previous_block_hash() == &(self.index.last_block_hash.unwrap()) {
```
clayrab, line 90: I think before we get here we shoudl check if the block's parent_hash is in our hashmap

src/blockchain.rs
```diff
@@ -39,20 +58,169 @@ impl Blockchain {
...
+                    // with the longest chain.
+                    //
+                    // The function also returns the new_chain and old_chain,
+                    // which are both `ChainFork`s. `ChainFork` is just a container
+                    // that collects `Blocks` for each chain, and calculates the cumulative burnfee
+                    // for each chain. This makes comparing the chains very simple, as well as
+                    // simplifying winding and unwinding the chain if the chain needs to be reorganized
+                    if let Some((shared_ancestor_index, new_chain, old_chain)) =
+                        self.fork_chains((&block).hash())
+                    {
```
clayrab, line 109: I don't want fork_chains in version 1. This is a lot of extra complication.

clayrab:
I think we can agree we've moved past this now :)

PR #58, opened 2021-05-28
Saito message cleanup

trevelyan:
First push to Saito-Rust repository.

This adds the contents of the /data/blocks directory to the .gitignore file so that we don't have blocks saved or added there

It also fixes SaitoMessage::NewBlock --> SaitoMessage::Block for consistency with Transaction.


clayrab:
LGTM

PR #59, opened 2021-05-28
return of the shashmap

trevelyan:
Minor change -- shashmap should be a generic key=>value data structure as opposed to a class to one use-case. We can instantiate it in ways that make it clear what the shashmap is being used to manage:

let utxoset = Shashmap::new();

The storage class Stephen is putting together will basically need to support saving blocks to disk and managing the UTXO set in this routing client (i.e. no need to database support) so it shouldn't be a problem to use it as the API contact point for updates to the UTXO set.

Need another reason? Hannes suggested I update my user.email and I'm admittedly curious if it makes a different in how this appears. So welcome back, Shashmap. It has been too long!

### DIFF
--- a/src/blockchain.rs
+++ b/src/blockchain.rs
@@ -1,5 +1,5 @@
 use crate::block::{Block, BlockHeader};
-use crate::utxoset::UTXOSet;
+use crate::shashmap::Shashmap;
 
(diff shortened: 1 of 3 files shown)


escapedcat:
:D you got to lizten to ze ruzt

clayrab (approved):

clayrab (approved):

PR #64, opened 2021-05-28
Transaction refactor

clayrab:
I want to get Transaction into a better state before we start using it more and a refactor like this becomes even more difficult. Already it has been quite difficult to get this change done.

I think it makes sense to pause for a moment and make sure we get these Types done as well as we can because it is going to save a lot of effort even in the near short-term and it will be very difficult to change in the future.

Each of the 3 Transaction types I've made here represent a Transaction as it changes while moving through the network.

The TransactionBody is the serializable parts of the transaction which are ultimately signed. We should make the signature and path into Option<..>s, but I think this is cleaner because we have more type safety this way. If the team is very much against this though, I'd say we can refactor again(remove TransactionBody and make sig and path into Option<sig> and Option<path>, but I'd argue that this is actually simple because we don't ever have to deal with Option/Some/None and interfaces can make it clear whether they expect a Transaction to be signed.

What is feel more strongly about is a separation of Transaction types which represent the Tranasction which gets serialized and put into blocks and sent over the wire vs the Transaction we have in runtime, mostly handled in blockchain.rs and shashmap.rs, which will have extra metadata attached to it like tx_id, block_id, etc. I'm not 100% sure this type will even be needed, but I've added it here for now in an attempt to make it clear that this type of data should not be included in Transaction.


src/block.rs
```diff
@@ -202,14 +202,14 @@ impl Block {
             // and the block id
             let mut current_sid = 0;
 
-            for slip in tx.outputs_mut().iter_mut() {
+            for slip in tx.body.outputs_mut().iter_mut() {
```
BearGuy, line 205: Small nitpick, but any reason we want accessors on the `body` instead of on `Transaction`?

src/transaction.rs
```diff
@@ -36,123 +58,96 @@ pub struct TransactionBody {
...
 }
 
-/// Enumerated types of `Transaction`s to be handlded by consensus
-#[derive(Debug, PartialEq, Clone)]
-pub enum TransactionBroadcastType {
-    Normal,
-}
-
 impl Transaction {
     /// Creates new `Transaction`
```
BearGuy, line 68: Docs need to be updated to reflect changes to function

BearGuy (commented):
I think the one downside to how this is implemented is that it's not easy to discern in the code when a transaction variable is a `TransactionBody` and when it's a `Transaction`. Perhaps we're more strict on variable naming between the two, but overall I see the incentive in separating them.

There's a tradeoff here between simplifying the `Transaction` struct and using multiple structs to get around serializing the wrong attributes, and using the type system to enforce `Transaction`s that can go in `Block`s. My feeling is that we can implement a custom `Serialize` trait to avoid serializing the `signature` in `TransactionBody`, so that doesn't require breaking `Transaction`s into multiple structs. As well, there's a question about what a type system enforcement on `Transaction` actually achieves. It stops one check of `Signature` to see if it contains a `None` signature in the `add_transaction` function in `Block`. Would there be other advantages to splitting out the struct, or is the only thing we get from it?

clayrab:
What is the tradeoff? I don't see a downside of extra types, it's a few extra lines of code which add semantic value and save many more lines of code... Can you try to explain to me what's the downside? I'm happy to rename the structs and variables and methods.

Having TransactionBody separate from Transaction achieves type safety and reduces compexity. no need to zero/null fields, no need to check if fields are zero/null, no need to make serialization extra complicated if we use bincode or protobufs. I'm happy to merge them if everyone disagrees, but I think the PR moves us forward and I'd like to merge it.

src/block.rs
clayrab: If I'm only dealing with a body(now called UnsignedTransaction), I'd still like to be able to get the outputs etc.

clayrab: we can wrap them if you really want to.

David and you are both pushing back against this "body" concept but that basically means we need to write a serialize() function instead, which in my view defeats the purpose of using bincode in V1, but I might just do it.

PR #66, opened 2021-05-28
clean up slip

clayrab:
The purpose of this is to fix shashmap in two PRs. I'm not sure what the proper data structures of shashmap, but we will have trouble discussing it I think until we get the Slip types right.

An output only needs an address and amount(and maybe a type). An input, is a reference to an output, which I'm refering to as "SlipID" these days. In saito-lite, we use (block_id, tx_id, slip_id) as SlipID. I don't think this is correct, but I want ot save that disucsion for another PR. The important thing to me is that we get these types right, i.e. Output = (address, amount, type) and Input = a reference/id. Then, I think the shashmap maybe only need to contain the Inputs/SlipIDs....

The current implementation mixes all this into a single type and thereby necessates zeroing/nulling out values at times, doesn't enforce type-safety when only one of the above types is really needed, makes serialization extra confusing, and potentially make the shashmap extra diffcult to maintain correctly and unnecessarily large.

### DIFF
--- a/src/shashmap.rs
+++ b/src/shashmap.rs
@@ -1,13 +1,12 @@
-use crate::slip::Slip;
+use crate::slip::{OutputSlip, SlipID};
 use crate::transaction::Transaction;
 use std::collections::HashMap;
 
-/// A hashmap storing the byte arrays of `Slip`s as keys
-/// with the `Block` ids as values. This is used to enforce when
-/// `Slip`s have been spent in the network
+/// A hashmap storing Slips TODO fix this documentation once we've settled on what
+/// data structures actually belong here.
 #[derive(Debug, Clone)]
 pub struct Shashmap {
-    utxo_hashmap: HashMap<Slip, i64>,
+    utxo_hashmap: HashMap<SlipID, OutputSlip>,
 }
 
 impl Shashmap {
@@ -29,10 +28,10 @@ impl Shashmap {
     /// * `tx` - `Transaction` where inputs are inserted, and outputs are removed
     pub fn unspend_transaction(&mut self, _tx: &Transaction) {}
 
-    /// Return the `Block` id based on `Slip`
+    /// Return the `Block` id based on `OutputSlip`
     ///
-    /// * `slip` - `&Slip` as key
-    pub fn slip_block_id(&self, slip: &Slip) -> Option<&i64> {
+    /// * `slip` - `&OutputSlip` as key
+    pub fn slip_block_id(&self, slip: &SlipID) -> Option<&OutputSlip> {
         self.utxo_hashmap.get(slip)
     }
 }
--- a/src/block.rs
+++ b/src/block.rs
@@ -194,28 +194,6 @@ impl Block {
 
     /// Sets the `Block`s list of `Transaction`s
     pub fn set_transactions(&mut self, transactions: &mut Vec<Transaction>) {
-        let bid = self.header.id;
-
-        for (i, tx) in transactions.iter_mut().enumerate() {
-            // The slips are assigned the ids based on their slip index
-            // in reference to all slips in the block, the transaction index,
-            // and the block id
-            let mut current_sid = 0;
-
-            for slip in tx.body.outputs_mut().iter_mut() {
-                slip.set_block_id(0);
-                slip.set_tx_id(0);
-                slip.set_slip_id(current_sid);
-                current_sid += 1;
-            }
-
-            for slip in tx.body.inputs_mut().iter_mut() {
-                slip.set_block_id(bid);
-                slip.set_tx_id(i as u64);
-                slip.set_slip_id(current_sid);
-                current_sid += 1;
-            }
-        }
         self.body.transactions = transactions.to_vec();
     }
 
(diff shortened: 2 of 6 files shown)


src/shashmap.rs
```diff
@@ -29,10 +28,10 @@ impl Shashmap {
     /// * `tx` - `Transaction` where inputs are inserted, and outputs are removed
     pub fn unspend_transaction(&mut self, _tx: &Transaction) {}
 
-    /// Return the `Block` id based on `Slip`
+    /// Return the `Block` id based on `OutputSlip`
     ///
-    /// * `slip` - `&Slip` as key
-    pub fn slip_block_id(&self, slip: &Slip) -> Option<&i64> {
+    /// * `slip` - `&OutputSlip` as key
+    pub fn slip_block_id(&self, slip: &SlipID) -> Option<&OutputSlip> {
```
BearGuy, line 34: This function name doesn't make any sense with the proposed changes. Should be something like `get_output_slip`

src/shashmap.rs
```diff
@@ -29,10 +28,10 @@ impl Shashmap {
     /// * `tx` - `Transaction` where inputs are inserted, and outputs are removed
     pub fn unspend_transaction(&mut self, _tx: &Transaction) {}
 
-    /// Return the `Block` id based on `Slip`
+    /// Return the `Block` id based on `OutputSlip`
```
BearGuy, line 31: Comment doesn't make sense under proposed changes. Should be "Return `OutputSlip` from `SlipID`"

BearGuy (commented):
How does the `Shashmap` scale when we are keeping track of `OutputSlip`s instead of regular `i64`s? The role of the `Shashmap` should be has been simple, enforcing existence of slips and not keeping track of any additional Slip data.

I don't have any issue with just using a`SlipID` as they key, as long as the `SlidID` regains a unique has to avoid collision.

trevelyan:
> An output only needs an address and amount(and maybe a type). An input, is a reference to an output...

I think there is some ambiguity in the way slips are being conceptualized here that flows from saito-lite "re-using" slips in the wallet. It is true that wallets have the concept of inputs and outputs. Also need to track additional information like LC and block HASH. We hack that stuff onto wallet-level slip storage just to simplify life in saito-lite (and it probably doesn't success and should be eliminated in the future).

Slips for consensus have no real concept of input / output. They just hold the data we need to get the answer to the question "is this spendable" and "does it belong to X" from the shashmap. Only the shashmap knows at any point in time if the slip is spendable.

The key is somewhat large (because it has to have the same security as the publickey), but the value it maps to is 0 or 1. That is one reason flicking through it so quickly is possible. In terms of easy wins here, as scalability depends on the size of both key and value to my understanding, I'd suggest we think about eliminating BID completely and shorten our key by adding tid / sid / amount together creatively (we will presumably have to make sure amount + tid can never rollover, lest someone claim a small payment in fact is quite a large one). Important thing is basically that the key is unique, so the slip is supposed to store this info. We could look at hashing it all together to reduce space, but any hash collisions would enable X to spend Y.

clayrab:
> > An output only needs an address and amount(and maybe a type). An input, is a reference to an output...
> 
> I think there is some ambiguity in the way slips are being conceptualized here that flows from saito-lite "re-using" slips in the wallet. It is true that wallets have the concept of inputs and outputs. Also need to track additional information like LC and block HASH. We hack that stuff onto wallet-level slip storage just to simplify life in saito-lite (and it probably doesn't success and should be eliminated in the future).

yes, this is the problem I'm trying to solve. Slips appear in 4 places: Outputs, Inputs, Utxoset, and Wallet. I think simply thinking of them as Outputs and OutputIDs is clearest, but I don't really care if we call the Slip and SlipID or Output and OutputID.
 
> Slips for consensus have no real concept of input / output. They just hold the data we need to get the answer to the question "is this spendable" and "does it belong to X" from the shashmap. Only the shashmap knows at any point in time if the slip is spendable.

I don't agree with this exactly... the presence of a Slip in the UTXOSet implies it's spendable, which would mean it's an unspent output. I'm not sure what point you're trying to make here. If we need both OutputSlip and SlipID in the utxoset, that's fine.

> The key is somewhat large (because it has to have the same security as the publickey), but the value it maps to is 0 or 1. That is one reason flicking through it so quickly is possible. In terms of easy wins here, as scalability depends on the size of both key and value to my understanding, I'd suggest we think about eliminating BID completely and shorten our key by adding tid / sid / amount together creatively (we will presumably have to make sure amount + tid can never rollover, lest someone claim a small payment in fact is quite a large one). Important thing is basically that the key is unique, so the slip is supposed to store this info. We could look at hashing it all together to reduce space, but any hash collisions would enable X to spend Y.

I'm 100% in agreement about elimited BID from SlipID, but I wanted to keep that as a separate conversation because I'm really just trying to get the Types right with this PR and I didn't want to cloud the conversation with a discussion about what is the right SlipID and how we implement the utxo set.

I"m also trying to walk a line between the words "Slip" and "Output", and frankly, I think the team's use of the word Slip is not helpful, even if it's somehow more technically correct, the industry has already settled on "output". Hence I'm using "OutputSlip" to try to keep things clear.

The point is there is something which is created by transactions(slips/outputs), and then those thing are referenced when they are spent. Whatever we want to call them, I don't care, but we need to be careful to distinguish their usage better than saito-lite does and it scares me when I see code setting the ID(block/tx/slip) to (0,0,0) and a need to check for Null/0 whenever we want to add slips to the UTXOset.

I think this moves us forward. Unless there are specific suggestions on downsides of doing things this way or changes that we need to make, I'd like to get this merged and add two more tasks, one to settled on what is the proper SlipID another to figure out shashmap's API and underlying datastructures.

PR #68, opened 2021-05-29
stub functions for epoch stack

clayrab:
I believe a data structure like I've laid out here is going to make the add_block algorithm much easier to write.

This is basically where we stock all the block_ids. The location of the hash in the array corresponds to the block_id. We can get latest block and the block at the bottom of the epoch.


clayrab:
We're merging this further up the commit tree

PR #71, opened 2021-05-30
Types refactor

clayrab:
This is getting merged and we're rebuilding from here.

This includes new types for Slip, Transaction, and Block. BlockIndex is removed. Mempool and golden_ticket have been removed. We will start from square 1 again with these.


src/blockchain.rs
```diff
@@ -1,116 +1,100 @@
...
+    // use crate::transaction::{Transaction, TransactionType};
+    // use secp256k1::Signature;
 
-        blockchain.add_block(block.clone());
-        let (block_header, _) = blockchain.index.blocks[0].clone();
-
-        assert_eq!(block_header, *block.clone().header());
-        assert_eq!(blockchain.shashmap.slip_block_id(&slip), Some(&-1));
-    }
+    // #[test]
```
escapedcat, line 52: Instead of commenting out tests you could use `ignore`:  
https://doc.rust-lang.org/reference/attributes/testing.html#the-ignore-attribute

escapedcat:
I'm confused by the amount commented out code. Is this needed?

src/blockchain.rs
clayrab: Okay, but they might not even build at the moment... I'll open up a new brnach and try to get things cleaned up today.

PR #75, opened 2021-06-04
feat: add_block method

clayrab:
This is a basic working add_block method and stubs for most related units(fork_tree, longest_chain_queue, utxoset, mempool).

Beside some questions about async/tokio, this code should be able to act as a platform to get to a functional testnet and should enable us to make a realistic roadmap.

In general, PRs this large are discouraged, but the complexity of add_block made it difficult to do incrementally. Stephen and I have worked closely over the past week to get this done but will strive to work in more bite-sized PRs going forward so that it's more easy for the community and team to engage in the process.


src/block.rs
```diff
@@ -49,10 +69,15 @@ impl Block {
     }
 
     /// Returns the `Block` hash
-    pub fn clone_hash(&self) -> [u8; 32] {
+    pub fn clone_hash(&self) -> Sha256Hash {
```
batzor, line 72: I think it's better and consistent if hash() returns &self.hash and if cloning is necessary, user should do block.hash().clone().

clayrab: This merge is messed up. hold on.

PR #78, opened 2021-06-04
Chore/convco

escapedcat:
This adds a Github action to check the commit format of pull-requests according to the format.  
The check will be required to pass if this gets merged.  
It will check a maximum of 50 commits on a PR. A PR shouldn't really have 50 anyways. We can change this later.

Also adding docs regarding the tooling.

### DIFF
--- a/.github/workflows/convco.yml
+++ b/.github/workflows/convco.yml
@@ -1,0 +1,21 @@
+name: Convco commit check
+# This workflow is triggered on pushes to the repository.
+on: [pull_request]
+
+jobs:
+  check:
+    name: Check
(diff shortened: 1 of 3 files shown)


clayrab:
What does this mean for commits made on a feature branch like clay_sandbox where I just want to save some random work or feature_branch_x where I'm going home for the day and would just like to commit code that's in a half-complete state?

clayrab:
LGTM.

Spoke with Hannes. This only applies to pull requests. I think this is good. I want to squash and merge for merges to main anyway, so this shouldn't get in the way of any of our current process. Let's try it out, we can also remove it.

PR #80, opened 2021-06-07
Burnfee and Block Validation Refactor

BearGuy:
Apologies for the massive PR, but this was split off from our `add_block` work and deserved a seperate PR.

In this PR, we
- Change `Block` to contain `difficulty` and `start_burnfee` inside `BlockCore` instead of at the high level
- Simplify how BurnFee operates, reintroduce `start_burnfee` for validation purposes
- Introduce different transaction validation for `Normal` and `GoldenTicket` transactions
- Simplify how `consensus` is running

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -44,7 +44,6 @@ impl Block {
             0,
             TREASURY,
             0.0,
-            0,
             0.0,
             transactions,
         );
@@ -60,10 +59,6 @@ impl Block {
         self.core.difficulty
     }
 
-    /// Returns the `Block` burnfee
-    pub fn burnfee(&self) -> u64 {
-        self.core.burnfee
-    }
     /// Returns the `BlockCore` of `Block`
     pub fn core(&self) -> &BlockCore {
         &self.core
@@ -157,8 +152,6 @@ pub struct BlockCore {
     treasury: u64,
     /// Start value in the `Burnfee` algorithm
     start_burnfee: f64,
-    /// `BurnFee` containing the fees paid to produce the block
-    burnfee: u64,
     /// Block difficulty required to win the `LotteryGame` in golden ticket generation
     difficulty: f32,
     /// simplified transaction cores
@@ -183,7 +176,6 @@ impl BlockCore {
             0,
             TREASURY,
             0.0,
-            0,
             0.0,
             &mut vec![],
         )
@@ -198,7 +190,6 @@ impl BlockCore {
         coinbase: u64,
         treasury: u64,
         start_burnfee: f64,
-        burnfee: u64,
         difficulty: f32,
         transactions: &mut Vec<Transaction>,
     ) -> Self {
@@ -210,7 +201,6 @@ impl BlockCore {
             coinbase,
             treasury,
             start_burnfee,
-            burnfee,
             difficulty,
             transactions: transactions.to_vec(),
         }
@@ -282,7 +272,6 @@ mod test {
         assert_eq!(block.previous_block_hash(), &[0; 32]);
         assert_eq!(block.creator(), public_key);
         assert_eq!(*block.transactions(), vec![]);
-        assert_eq!(block.burnfee(), 0);
         assert_eq!(block.difficulty(), 0.0);
         assert_eq!(block.coinbase(), 0);
         assert_eq!(block.start_burnfee(), 0.0);
--- a/src/block.rs
+++ b/src/block.rs
@@ -6,16 +6,14 @@ use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 use std::{mem, slice};
 
+pub const TREASURY: u64 = 286_810_000_000_000_000;
+
 /// The `Block` holds all data inside the block body,
 /// and additional metadata not to be serialized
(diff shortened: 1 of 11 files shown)


trevelyan:
Clay explicitly asked me to hold off for this push and I need to be working from an updated version. I do not think the issues with the code in its proposed state is any worse than before, and there are improvements. So I am in favour of this merge.

There are issues here with burnfee -- we should not need the start_burnfee and burnfee as two separate consensus variables -- the current code only has a single one as I believe (we used to have two) as we can get the earlier of the two figures during validation from the previous block, which needs to be accessible as well in order to calculate monetary policy, etc.

From looking at the amount of code that is needing to be changed, it feels very much like the code is moving ahead of where its fundamentals support, so that simple corrections are taking quite a bit of time. I will try to put some effort into thinking through this.

src/block.rs
```diff
@@ -146,6 +153,14 @@ pub struct BlockCore {
     creator: PublicKey,
     /// Total block reward being released in the block
     coinbase: u64,
+    /// Amount of SAITO left in reserve on the network
+    treasury: u64,
+    /// Start value in the `Burnfee` algorithm
+    start_burnfee: f64,
```
clayrab, line 159: Sorry if we keep repeating conversations, but can't I get the start_burnfee from previous block's burnfee?

src/blockchain.rs
```diff
@@ -217,86 +219,131 @@ impl Blockchain {
...
-        if block.burnfee()
-            != BurnFee::burn_fee_calculation(
-                previous_block.burnfee() as f64,
+        if block.start_burnfee()
+            != BurnFee::burn_fee_adjustment_calculation(
+                previous_block.start_burnfee() as f64,
                 block.timestamp(),
                 previous_block.timestamp(),
-            )
+            ) as f64
```
clayrab, line 235: The previous code makes more sense to me, I still don't see why we need start_burnfee...

src/blockchain.rs
```diff
@@ -217,86 +219,131 @@ impl Blockchain {
...
+
+        let golden_ticket_count = block.transactions().iter().fold(0, |acc, tx| {
+            if tx.core.broadcast_type() == &TransactionType::GoldenTicket {
+                acc + 1
+            } else {
+                acc
+            }
+        });
+        if golden_ticket_count > 1 {
+            println!("TOO MANY GOLDEN TICKETS");
```
clayrab, line 256: WhyCantIHoldAllTheseLimes.gif

src/blockchain.rs
```diff
@@ -217,86 +219,131 @@ impl Blockchain {
...
             .iter()
-            .all(|tx| self.validate_transaction(block, tx, fork_tuple));
+            .all(|tx| self.validate_transaction(previous_block, block, tx, fork_tuple));
 
         transactions_valid
     }
 
     fn validate_transaction(
         &self,
+        previous_block: &Block,
```
clayrab, line 270: Do we only need the previous block hash? Can we just get it from block?

src/blockchain.rs
```diff
@@ -217,86 +219,131 @@ impl Blockchain {
...
-                return false;
+                if let Some(output_slip) =
+                    self.utxoset.output_slip_from_slip_id(&tx.core.inputs()[0])
+                {
+                    let hash_message = tx.core.hash();
+                    if !verify_message_signature(
+                        &hash_message,
+                        &tx.signature(),
+                        &(output_slip.address()),
+                    ) {
```
clayrab, line 289: It's not immediately obvious that we need to get an input in order to validate the signature. I think some comment here could be very helpful. There is a stub function in utxo_set called get_receiver_for_slips. Please use that instead of output_slip_from_slip_id. The idea is that all the inputs need to have the same receiver and utxoset has this information. if it returns None, then the tx is invalid. This also avoids the confusion of "output_slip" caused because output_slip is a bit confusing to read here because it's an output from another transaction.

src/blockchain.rs
```diff
@@ -217,86 +219,131 @@ impl Blockchain {
...
+                        println!("SIGNATURE IS NOT VALID");
+                        return false;
+                    };
+
+                    // validate our slips
+                    let inputs_are_valid = tx.core.inputs().iter().all(|input| {
+                        return !self
+                            .utxoset
+                            .is_slip_spent_at_block(input, block, fork_tuple);
+                    });
```
clayrab, line 299: hmmm, there must be a nicer way to do this... 

Can we match on this and reduce it to just sticking a "?" on the end and get rid of inputs_are_valid?

Happy to accept this as-is, but maybe a little "TODO" would be nice to acknowledge that this might not be the rustiest way to do it.

clayrab:
Please just address the few issues I've mentioned and then feel free to merge.

We need to sort out the consensus.rs and how we're going to use tokio, but I don't want to block this if I don't have a solution to that and we need some async timeout to make burnfee work at all, so in the interest of moving things forward, let's just merge this and then sort out the tokio/async/memory-sharing stuff ASAP.

src/block.rs
BearGuy: Nope, it's the opposite! You get the current burnfee form the previous block's `start_burnfee`

PR #84, opened 2021-06-09
fix: documentation updates.

clayrab:
Update docs


README.md
```diff
@@ -35,7 +34,11 @@ cargo run
 ### Tests
 
 ```
-cargo test
+scripts/test.sh
+```
+or 
+```
+cargo test --features=test-utilities
```
escapedcat, line 41: Can you add why this is needed?

escapedcat:
@clayrab if you only change docs you can use `docs: ...` instead of `fix: ...`  
We don't need a new patch version because of this in the future.

PR #86, opened 2021-06-10
First Networking Implementation

BearGuy:
This is the first networking implementation which supports a basic http server. We use warp to expose an HTTP endpoint which serves binary blocks over the network based off of the block hash


README.md
```diff
@@ -34,11 +35,7 @@ cargo run
 ### Tests
 
 ```
-scripts/test.sh
-```
-or 
-```
-cargo test --features=test-utilities
+cargo test
```
escapedcat, line 38: Is this correct or is a rebase missing?

BearGuy: `test_utilities` is not hidden behind a feature flag currently, so it's not necessary to declare it

clayrab:
We want to finish up the networking later, but this is useful code and it's becoming a big chore to keep it synced with main. We're going to merge this for now.

PR #87, opened 2021-06-10
Benchmarking Framework with benchmarking `add_block`

BearGuy:
Introducing criterion to run our benchmarking so we can start to get some hard data on how quickly we can test add blocks to the blockchain.


src/blockchain.rs
```diff
@@ -326,23 +326,24 @@ impl Blockchain {
                     let output_amt: u64 =
                         tx.core.outputs().iter().map(|output| output.amount()).sum();
 
-                    if input_amt != output_amt {
+                    if input_amt < output_amt {
```
clayrab, line 329: I think this should be !=

PR #90, opened 2021-06-11
Utxoset

clayrab:
Utxoset. This handles validating spendability on both longest chain and potential forks.

There are still many optimizations which we are planning to do here, but this should be a good start to get a working node and something to benchmark against for future optimization work.

Tests still need more work.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,9 +1,0 @@
-[env]
-EPOCH_LENGTH = 1000
- 
-[rustc-env]
-EPOCH_LENGTH = 1000
-# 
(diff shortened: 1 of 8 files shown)


escapedcat:
@BearGuy just nitpicking:  
image  
- It shoudl be `feat` not `feature`
- If you keep the subject (message) of your commit a maximum of 70 chars it will not be cut off  
  Every info longer than 70 chars can go into the commit-`body`

Depending on how you create your commits there are different ways to add info to the body. I.e. like this:

```
git commit -m 'feat: such a nice thing

Here goes more info into the body'
```

PR #91, opened 2021-06-11
fix: github tests need proper configuration

clayrab:
Github needs the proper test config


clayrab:
fix: github has wrong test config

escapedcat:
Can you describe why this is needed (in general). Could add it to the readme.

PR #94, opened 2021-06-12
chore: add clippy code linting

escapedcat:
## Background

While looking through the code I saw this example:  https://github.com/SaitoTech/saito-rust/blob/main/src/blockchain.rs#L140-L189  

A deeply nested if/else block like this wouldn't pass any [linting in the JS world](https://eslint.org/). So I thought there must be something similar in the Rust world. I found [Clippy](https://github.com/rust-lang/rust-clippy) which seems to be popular.  

## Current status

[Running clippy in the current `main` branch we get](https://github.com/SaitoTech/saito-rust/pull/94/checks?check_run_id=2808754792) (you might need to click `clippy` after opening this link):

```
error: aborting due to 3 previous errors; 87 warnings emitted

error: could not compile `saito_rust`
```

This might not be too bad actually. The `errors` should be fixed. The `warnings` are a nice-to-have.  
The specific rules can be adjusted.   

## Intergrating Clippy into the project as a Github action

I think this can help to keep code in a good state and as Clippy states:  

> to catch common mistakes and improve your Rust code

This PR would add it to the Github actions, so we will see a ❌. It's non-blocking though (for now)!

Are we in?!

P.s.
- Here is a short video which shows Clippy helping with two examples:  
  [Rust For Beginners Tutorial - Linting with Clippy](https://www.youtube.com/watch?v=BE9KY6X7aUM&list=PLauX9TuJ8sfyaLPZ1udS3zS_V9YXdsbtc)
- [The Rust Performance Book about Clippy](https://nnethercote.github.io/perf-book/linting.html)


clayrab:
Can't say I'm a fan of this one... add_block is a necessarily complex function and I think it would actually make it more complex to break it apart. So in this particular case clippy would just be annoying... Is there another reason we need it given that we already are using cargo fmt?

escapedcat:
@clayrab I understand what you mean. Linting (clippy) and formatting (rustfmt) code are too different things.  
Let's talk about it in person. Easier to explain why this is useful and why you don't have to worry :)

escapedcat:
Apart from warnings these are the current linting errors:

```bash
error: you are implementing `Hash` explicitly but have derived `PartialEq`
  --> src/slip.rs:15:1
   |
15 | / impl Hash for SlipID {
16 | |     fn hash<H: Hasher>(&self, state: &mut H) {
17 | |         state.write(&self.slip_ordinal.to_ne_bytes());
18 | |         state.write(&self.tx_id)
19 | |     }
20 | | }
   | |_^
   |
   = note: `#[deny(clippy::derive_hash_xor_eq)]` on by default
note: `PartialEq` implemented here
  --> src/slip.rs:23:58
   |
23 | #[derive(Serialize, Deserialize, Debug, Clone, Copy, Eq, PartialEq)]
   |                                                          ^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

```bash
error: this comparison involving the minimum or maximum element for this type contains a case that is always true or always false
   --> src/longest_chain_queue.rs:101:12
    |
101 |         if self.longest_chain_length <= 0 {
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[deny(clippy::absurd_extreme_comparisons)]` on by default
    = help: because `0` is the minimum value for this type, the case where the two sides are not equal never occurs, consider using `self.longest_chain_length == 0` instead
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#absurd_extreme_comparisons
```

```bash
error: strict comparison of `f32` or `f64`
   --> src/blockchain.rs:281:12
    |
281 |           if block.start_burnfee()
    |  ____________^
282 | |             != BurnFee::burn_fee_adjustment_calculation(
283 | |                 previous_block.start_burnfee() as f64,
284 | |                 block.timestamp(),
285 | |                 previous_block.timestamp(),
286 | |             ) as f64
    | |____________________^
    |
    = note: `#[deny(clippy::float_cmp)]` on by default
    = note: `f32::EPSILON` and `f64::EPSILON` are available for the `error_margin`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#float_cmp
help: consider comparing them within some margin of error
    |
281 |         if (block.start_burnfee() - BurnFee::burn_fee_adjustment_calculation(
282 |                 previous_block.start_burnfee() as f64,
283 |                 block.timestamp(),
284 |                 previous_block.timestamp(),
285 |             ) as f64).abs() > error_margin
    |
```

What do you think about these @BearGuy ?

BearGuy:
Looks like this would help keep code quality high, which I'm definitely for. As long as it's not too pedantic and isn't too cumbersome, I'm more than happy to add it

escapedcat:
@BearGuy I think it would be an improvement if we only take care of errors and let warnings be warnings for now. You think you can fix the 3 current errors?

BearGuy:
Due to the new rebase, I think it's out of scope to address the breaking changes in this PR. I'm going to merge first, then fix these changes in a different PR addressing what's missing

PR #95, opened 2021-06-14
Refactor: remove extraneous code and update tests

BearGuy:
Just adding some additional tests and clean up some functions and comments.


clayrab:
LGTM. I didn't like the name of crypto::hash() so I renamed it to hash_bytes and it was duplicate functionality of make_message_from_bytes, so I got rid of that also.

PR #97, opened 2021-06-15
Feature: Transaction fee calculation and Routing fee calculation with validation

BearGuy:
This PR calculates payable fees in each transaction, as well as designating what fees can be collected based on the amount of hops contained in a transaction.


LICENSE
```diff
@@ -1,21 +0,0 @@
-MIT License
```
escapedcat, line 1: So no more license at all? :(

BearGuy: We can't have MIT license if we're including code that's linked to IP. Might be ok, but being a bit cautious

escapedcat: Sure, I understand. Would be great to decide for a license because otherwise people won't know if and how to use this.  
/cc @arpee

PR #114, opened 2021-06-23
Feature: mempool run with message passing and try bundle

BearGuy:
Wanted to address some issues that I saw in the initial mempool.

My preference with regards to running mempool is through channels rather than Arced variables. The channel acts as a queue to apply changes and command the mempool to run certain operations, like bundling a block or adding a transaction.

With channels, we're still able to operate on mempool across different threads without concern about data races or state changes occuring in the middle of an operation.


trevelyan:
Happy with the general changes -- will merge.

Still not convinced we want a systems broadcast channel to be serving the purpose of a loop. But the execution is cleaner, and the delineation of functions provides a cleaner separation of logic. It will be easier to shift this to managing block creation when we start basing block creation more off routing work.

Minor feedback:

1. Like the idea of ::run() functions that setup generic receivers within these classes.

2. The Mempool will need the broadcast_sender re-included in its constructor so that the mempool can broadcast messages in the course of normal operation (block added, transaction added, mempool congested, etc.).

3. can_bundle --> can_bundle_block ---- lets make all can_* functions mirror the full name of the function they are gatekeeping.

trevelyan:
Note also -- getting the mempool.write() request as contained as possible so we don't make write requests on every message,etc.


  while let Ok(message) = mempool_channel_receiver.recv().await {
        match message {
            let mut mempool = mempool_lock.write().await;
            SaitoMessage::TryBundle => {
                if mempool.can_bundle(blockchain_lock.clone()) {
                    mempool.bundle_block(blockchain_lock.clone());
                }
            }
        }

PR #116, opened 2021-06-24
updated classes info

trevelyan:
This adds block and transaction and slip classes. Also the time class as that is needed for mempool to function.

Each of them has a Core in which we can store the data that needs to be saved to blocks or shared with peers. The full object can then contain other information, like the state of validation, or generated hashes, etc.

I know that Clay is working on serialization. Have still done this as it gives us the option of using Bincode to bootstrap things in the short-term. The intent is to have these objects as private variables so that we can eliminate them in the future once manual serialization is possible.

All data objects here are byte arrays for speed with serialization and deserialization. The only time that we should need to move TX and SLIP objects out of this type is during validation, which happens in parallel.


clayrab:
LGTM

PR #119, opened 2021-06-24
Sequential refactor adding constant crypto types

trevelyan:
Transactions, Blocks and Slips refactored to be human-readable, using types defined in crypto.rs:

pub type SaitoHash = [u8; 32];
pub type SaitoPublicKey = [u8; 33];
pub type SaitoSignature = [u8; 64];

I'm hoping that we can keep manipulations of these items restricted to the serialization functions and the cryptographic functions. Would be nice to have these setup to provide default zero'd-out values, but I don't think this is possible in Rust with aliases.

We are probably not going to want to change cryptographic primitives, but having this all in the crypto class is probably a positive in keeping cryptographic complexity contained.


trevelyan:
Have made some additional changed pushed here that get the mempool publishing block notifications and the blockchain class taking ownership of the blocks (at which points they are deleted -- we don't have block storage yet). This basically links the work Stephen did last night with the actual Blockchain and Mempool classes.

PR #120, opened 2021-06-24
wallet class and keypair in crypto class

trevelyan:
This creates a generic wallet that works exactly like the mempool.
 - private channel for in-wallet messages
 - ability to listen on and public system-wide broadcasts
 
This pushes the Keypair INTO the crypto-class as part of our efforts to contain anything that touches algorithm-specific code to that class. The wallet has a copy, but gets_ the standard bytes-array data. We may want to think about not storing the Keypair in the wallet at all.

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -1,5 +1,6 @@
 use crate::{blockchain::Blockchain, mempool::Mempool};
 use crate::crypto::{SaitoHash};
+use crate::wallet::Wallet;
 use std::{future::Future, sync::Arc};
 use tokio::sync::RwLock;
 use tokio::sync::{broadcast, mpsc};
@@ -11,6 +12,7 @@ struct Consensus {
     _shutdown_complete_rx: mpsc::Receiver<()>,
     _shutdown_complete_tx: mpsc::Sender<()>,
 }
+
 
 /// The types of messages broadcast over the main
 /// broadcast channel in normal operations.
@@ -70,6 +72,7 @@ impl Consensus {
         //
         let blockchain_lock = Arc::new(RwLock::new(Blockchain::new()));
         let mempool_lock = Arc::new(RwLock::new(Mempool::new()));
+        let wallet_lock = Arc::new(RwLock::new(Wallet::new()));
 
         tokio::select! {
             res = crate::mempool::run(
@@ -83,6 +86,15 @@ impl Consensus {
                 }
             },
             res = crate::network::run(
+                broadcast_channel_sender.clone(),
+                broadcast_channel_sender.subscribe()
+            ) => {
+                if let Err(err) = res {
+                    eprintln!("{:?}", err)
+                }
+            }
+            res = crate::wallet::run(
+                wallet_lock,
                 broadcast_channel_sender.clone(),
                 broadcast_channel_sender.subscribe()
             ) => {
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -1,22 +1,62 @@
-pub use secp256k1::{Message, PublicKey, Signature, SECP256K1};
+pub use secp256k1::{Message, SecretKey, PublicKey, Signature, SECP256K1};
 use sha2::{Digest, Sha256};
 use std::convert::TryInto;
-
+use base58::ToBase58;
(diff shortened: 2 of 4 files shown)


src/consensus.rs
```diff
@@ -70,6 +72,7 @@ impl Consensus {
         //
         let blockchain_lock = Arc::new(RwLock::new(Blockchain::new()));
         let mempool_lock = Arc::new(RwLock::new(Mempool::new()));
+        let wallet_lock = Arc::new(RwLock::new(Wallet::new()));
```
clayrab, line 75: Does wallet need to be in a mutex?

src/consensus.rs
```diff
@@ -90,6 +93,15 @@ impl Consensus {
...
             }
+            res = crate::wallet::run(
+                wallet_lock,
+                broadcast_channel_sender.clone(),
+                broadcast_channel_sender.subscribe()
+            ) => {
+                if let Err(err) = res {
+                    eprintln!("{:?}", err)
+                }
+            }
```
clayrab, line 104: What is the purpose of wallet run function? it seem to be an initialization, but I'm not clear why it's done this way.

src/crypto.rs
```diff
@@ -1,23 +1,63 @@
...
+    pub fn get_publickey(&self) -> SaitoPublicKey {
+	self.publickey.serialize()
+    }
+
+}
+
 pub fn hash(data: &Vec<u8>) -> SaitoHash {
     let mut hasher = Sha256::new();
     hasher.update(data);
     hasher.finalize().as_slice().try_into().unwrap()
```
clayrab, line 57: Should this be in the crypto namespace?

src/wallet.rs
```diff
@@ -0,0 +1,146 @@
...
+
+
+
+
+
+//
+// This function is called on initialization to setup the sending
+// and receiving channels for asynchronous loops or message checks
+//
+pub async fn run(
```
clayrab, line 62: I'm confused why this function is needed or why wallet needs a loop or a recv().

clayrab:
This doesn't build on my machine.

error[E0277]: the trait bound `[serde_with::Same; 64]: DeserializeAs<'_, _>` is not satisfied
   --> src/transaction.rs:48:1
    |
48  | / #[serde_with::serde_as]
49  | | #[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
    | |_______________________________^ the trait `DeserializeAs<'_, _>` is not implemented for `[serde_with::Same; 64]`
    | 
   ::: /home/clay/.cargo/registry/src/github.com-1ecc6299db9ec823/serde_with-1.9.4/src/lib.rs:373:12
    |
373 |           T: DeserializeAs<'de, I>,
    |              --------------------- required by this bound in `As::<T>::deserialize`
    |
    = help: the following implementations were found:
              <[As; 0] as DeserializeAs<'de, [T; 0]>>
              <[As; 10] as DeserializeAs<'de, [T; 10]>>
              <[As; 11] as DeserializeAs<'de, [T; 11]>>
              <[As; 12] as DeserializeAs<'de, [T; 12]>>
            and 29 others
    = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to 12 previous errors

src/consensus.rs
trevelyan: I think these are positive changes. Happy to switch to Mutex as well as above. Rationale for an independent class is that if wallet can listen on OnNewBlock or OnChainReorganization and then fetch the slips or update the spendable ones instead of being manually triggered by the blockchain class that would be an improvement.

Right now this class mostly exists as a place to store the Keypair so that we can get the publickey and privatekey into the mempool for transaction / block creation and signing in order to get blocks and transaction signature creation and validation worked on.

src/crypto.rs
trevelyan: goal is keeping all of the crypto-related functions in one place, and have them operate on base data structures.

hash / sign / validate

aim is to avoid inconsistencies in how external classes hold cryptographic data objects, and having a larger number of these functions that sprout up with different input arguments depending on degree of initialization etc. only 1 function to test.

PR #121, opened 2021-06-24
Unit tests and Default trait impl for base structs, and proposed changes to Mempool operations

BearGuy:
Apologies for this one, I mismanaged my work and mixed in two different initiatives.

The first set of changes is more basic. The `Default` trait is implemented in our base structs where it's convenient, along with unit tests and some cleanup to satisfy our newly reintroduced checks.

The other staged changes are based around `Mempool` operations. I've restructured some things based on this list of assumptions:

1. Global channels are used to pass messages externally between different `run` functions. They shouldn't be used internally because those channels broadcast messages are replicated across many channels, clogging up the message queue.

2. A node should prioritize adding a bundled block to the longest chain over any other external block. This means that a bundled block should be expedited into the chain through implementing its own unique `add_block` path.

3. It should be decided concretely whether structs like `Mempool` and `Blockchain` should have local access to the global broadcast channel instead hedging and have them implemented as an `Option`. Personally, I prefer to keep channel operations inside the `run` function and leave a structs implemented functions to purely get and set information, but I'm open to discuss this further.

A few other things
- `Mempool` now has an `AHashMap<SaitoHash, Block>` for `blocks` instead of a `Vec<Block>`. This radically simplifies adding and taking blocks. This simplicity can be seen in `add_block` and `take_block` (formally `get_block`).
- Mempool's `get_block` has been renamed to `take_block` to have it differentiated from functions that merely grab a reference to data. `take_block` is different because it mutates the state of `Mempool` and transfers ownership of the `Block` data, which means that it should have a different naming scheme than `get_`. 
- I've tried to simplify `can_bundle_block` and `bundle_block` by passing in a reference to the latest block rather than passing in the guarded blockchain. Depending on the complexity of `bundle_block` in the future, this could need to be reverted. My intention is in the meantime to simplify functions on our main structs and handle guard complexity in our `run` functions. Again, open to discussion on this.


src/mempool.rs
```diff
@@ -20,225 +17,137 @@ pub enum MempoolMessage {
...
-            }
-        }
-
-	self.blocks.push(block);
-        return true;
-
+    pub fn add_block(&mut self, block: Block) {
+        // adds block if it doesn't already exist,
+        // else it ignores attempt to add block to AHashMap
+        self.blocks.entry(block.get_hash()).or_insert(block);
```
clayrab, line 34: My intuition is that methods like this can cause bugs. If we're adding a block, it is better to know whether it needs to be added or not. If we are calling "add_block" for a block that has already been added, I think this is an indication that the calling code is unsure what it's doing. There may be special cases where you'd want this kind of "try" behaviour, but they are rare. Since we're not calling this function yet anywhere, it's hard to evaluate. I'd rather just insert or panic for now.

src/mempool.rs
```diff
@@ -20,225 +17,137 @@ pub enum MempoolMessage {
...
-
-	if block_found {
-	    let block = self.blocks.remove(block_idx);
-	    return Some(block);
-	}
-
-	return None;
-
+    pub fn take_block(&mut self, hash: SaitoHash) -> Option<Block> {
+        self.blocks.remove(&hash)
```
clayrab, line 38: nice

src/mempool.rs
trevelyan: A sanity check on not adding is presumably against:

 - someone generates a block with a hash collisions (improbable but it can happen -- tough luck, your candidate block is invalid)
 - some peculiar race condition it might be hard to spot where a duplicate block gets into the mempool queue while the first block is being processed but after it has left the mempool?
 
I agree with Clay that the design should not be passing blocks to add_block() unless we are sure we want to add them. Personally suspect the filtering should primarily happen in peer / mempool. Perhaps we should avoid the "ignores attempt" uncertaintly by explicitly sending the block to addBlocktoBlockchainFailure (or whatever the equivalent should be in Rust)?

trevelyan:
> A node should prioritize adding a bundled block to the longest chain over any other external block. This means that a bundled block should be expedited into the chain through implementing its own unique add_block path.

Should it, though? 

In many cases it is more in my interest to build upon it.

Let's keep things simple and focus on single-path block processing until that works well.

> Mempool now has an AHashMap<SaitoHash, Block> for blocks instead of a Vec<Block>. This radically simplifies adding and taking blocks. 

We need a queue for this data-structure, although there can be multiple parts (perhaps hashmap for storage and something else for handling order). Went with the Vec<Block> temporarily as the structure should not get very large. The code ended up being larger because you can't remove elements while iterating through a structure in Rust as far as I can tell. 

> I've tried to simplify can_bundle_block and bundle_block by passing in a reference to the latest block rather than passing in the guarded blockchain.

We need access to the blockchain -- in the future we will need the ATR-rebroadcast block. Possibly also any data structures associated with the blockchain like any ancillary stakers table that will need to live there as well.

> It should be decided concretely whether structs like Mempool and Blockchain should have local access to the global broadcast channel instead hedging and have them implemented as an Option. Personally, I prefer to keep channel operations inside the run function and leave a structs implemented functions to purely get and set information, but I'm open to discuss this further.

This is an important discussion. Our most important?

I suspect that for practical purposes it is going to boil down to how deeply chained the functions are once we call them from the message-trigger-points? If we end up with deeply-chained functions then we need to be passing these objects sequentially through them and it defeats the point of having a flexible class-structure.

For now, I personally think having run set_X is positive, as with deeply-chained functions we will need to pass all of the needed locks down with every stop and end up with function-argument-paralysis. 

We can probably use a local loop to create blocks. That at least avoids the need for a local copy of the wallet, reducing the question to handling of the blockchain_lock.

I'm not sure what is best for TDD. Curious what Clay thinks.

trevelyan (commented):
merging.

will make some changes to mempool

happy to work with keeping stuff in the run loop as long as possible.

PR #122, opened 2021-06-25
Sequential refactor crypto updates

trevelyan:
This is mostly improvements to the crypto class -- getting it so that it can sign messages and validate them.


escapedcat:
you can avoid failing the commit check by stickign to this commit format:  https://github.com/SaitoTech/saito-rust/blob/main/CONTRIBUTING.md#commit-format

i.e. `update: crypto signing function enabled` could be `feat: crypto signing function enabled`. `update` is not a valid type.

PR #123, opened 2021-06-25
fix: de/serialize big arrays

batzor:
Fixed build by implementing de/serialize for big arrays.
`#[serde_as(as = "[_; 33]")]` does not work when the array size is larger than 32 so we have to manually implement it.


escapedcat:
Thanks @batzor !

BearGuy (approved):
LGTM. I'll update the branch and get it merged here

PR #124, opened 2021-06-25
Refactor: Orienting Mempool operations around block queue with consistency

BearGuy:
My last changes didn't respect `Mempool` as a block queue well enough. I have no issue with this, but if we're going to have a block queue, it's best that we're consistent with a queue operations.

This means
- `Mempool` uses `VecDeque` as a queue. It has the ability to add unique blocks and process them FIFO.
- Blocks are added to the block queue, then added to the chain sequentially via a `ProcessBlocks` message.
- To adhere to a FIFO queue, there should be no function that attempts to prioritize blocks over others via `take_block`. We simply process the queue each time we wish to add a block to the chain.

There are still some questions to be answered
- How do different parts of the program add blocks into the queue? For now, my assumption is that we're passing the mempool reference, adding blocks via `add_block` then sending the message `MempoolNewBlock` to internally process the block queue.

### DIFF
--- a/src/mempool.rs
+++ b/src/mempool.rs
@@ -7,13 +7,19 @@ use crate::{
     transaction::Transaction,
     wallet::Wallet,
 };
-use std::{sync::Arc, thread::sleep, time::Duration};
+use std::{collections::VecDeque, sync::Arc, thread::sleep, time::Duration};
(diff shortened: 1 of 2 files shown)


trevelyan (approved):
This looks great!

PR #125, opened 2021-06-25
Refactor: demoted Wallet with limited access to private key 

BearGuy:
A few thoughts:

- Wallet can be passed as a local attribute for both `Mempool`, `Blockchain` and other parts of the program
- It does not need a `run` function, all write operations can be done within `Blockchain`
- Access to `privatekey` should be hidden within `Keypair` inside `Wallet`. It's a security concern to pass around a node's private key all over the program
- Therefore, `sign` should be moved out of crypto and strictly into `Wallet`. The access problem is solved by passing `Wallet` as a reference widely to parts of the program that need it

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -68,15 +68,14 @@ impl Consensus {
         // whether the channels exist and unwrapping them when sending
         // messages, as well as setters.
         //
-        let blockchain_lock = Arc::new(RwLock::new(Blockchain::new()));
-        let mempool_lock = Arc::new(RwLock::new(Mempool::new()));
         let wallet_lock = Arc::new(RwLock::new(Wallet::new()));
+        let blockchain_lock = Arc::new(RwLock::new(Blockchain::new()));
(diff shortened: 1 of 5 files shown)


trevelyan:
I'm happy with the wallet_lock shift and getting rid of the run function. We can bring that back as and if we need it, but there's no reason for it at this point and this is cleaner and simpler. So thumbs up there.

Personally believe we should move in the other direction with the Keypair (eliminate it completely). The object is an artifact from the secp251k library rather than a cryptographic primitive. The only reason it existed at all in the wallet class was that it made generating keys easier.

> It's a security concern to pass around a node's private key all over the program

Whatever has access to the wallet by definition has access to the privatekey.

We should absolutely have a sign() function in the wallet. It can have complicated arguments and take conditional actions. But the actual cryptographic signing remains in crypto.js where it is blackboxed and can be upgraded or swapped out as easily as the hash function.

trevelyan (approved):
Mostly good -- will take a stab at the keypair stuff but don't want this holding us up.

PR #126, opened 2021-06-25
Feature: use blake3 for hashing algorithm with parallel hashing

BearGuy:
These numbers speak for themselves:

Benchmark

Orders of magnitude increase in speed without any major differences in security in comparison to sha2.

Additionally, blake3 allows us to hash in parallel. Based on their own benchmarks, inputs over 128 kb show improvements in speed when parallelized. This would be excellent in speeding up signature creation/verification for `Transaction`s

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,6 +21,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "arrayref"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a4c527152e37cf757a3f78aae5a06fbeefdb07ccc535c980a3208ee3060dd544"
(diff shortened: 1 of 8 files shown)


trevelyan (approved):
Have done some reading on Blake3 and think this is a good idea. The trade-off is supposed to be increased memory usage that might make the algorithm not suitable for lite-devices processing larger files. We shouldn't have that problem as non-full-nodes do not need to perform hashing.

Expected lifetime is similar to or better than SHA256 as far as a morning of research suggests, and -- as Stephen says -- the improvements speak for themselves. Merging in order to be able to work off this codebase. Will attempt to run some numbers against SHA256 this weekend to see if we can get any empirical confirmation of improvements.

PR #128, opened 2021-06-27
style: cargo fmt

escapedcat:
get `main` ✅


escapedcat:
Might just be resovled by #129

PR #129, opened 2021-06-27
Sequential refactor blockring and blockchain add block

trevelyan:
Will be continuing to work on this, but sharing a push so that people can see this code in advance. This brings the essentials of wind / unwind chain into Rust. It updates the Blockchain file so that blocks are processed and added in a chain. And it preserves but modifies the fast-access VecDeque that Clay has suggested as a BlockRing.

Blockring.rs is an abstract class and the exact data structures inside should be discussed and can be upgraded without affecting the logic of how we process blocks. The need for the elements in the Blockring to have variable-size components means we can no longer store this on the stack in contiguous memory, so I am wondering if the best approach might be using a SliceDeqVeque -- https://docs.rs/slice-deque/0.3.0/slice_deque/. We could also use a Vector and simply maintain a slice of entries in the struct that remains valid over the lifetime of the program and use them for fast access to elements in the middle if needed. Suggestions welcome.

In terms of functionality, the major change with this implementation is that the BlockRing maintains information on multiple blocks and chains as opposed to simply the hashes that form the longest-chain. It updates its own variable indicating which is the longest-chain when the chain is reorganized (on_chain_reorganization).This view into how many blocks exist at arbitrary block_ids is needed for ATR deletion and will be useful for congestion policy. It also avoids the need for the tree of blocks stored in the hashmap to contain information on their children. The addition of parentless blocks is also now possible. The data structure is more flexible.

Blockchain has gained AHashMap named "utxoset" that is serving as the UTXOSet. We have continued the tradition of using add_block() and on_chain_reorganization() as our two essential functions for adding blocks and updating state when the chain is updated. Prevalidation is possible and can permit us to skip wind / unwind chain step, although we now skip ALL interactions with the UTXOset if the block is not identified as a suitable candidate for the longest-chain.

We should be able to move forward fleshing out the validation steps and adding unit tests.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -1,10 +1,11 @@
 use crate::{
     big_array::BigArray,
-    crypto::{hash, SaitoHash, SaitoPublicKey, SaitoSignature},
+    crypto::{hash, SaitoHash, SaitoPublicKey, SaitoSignature, SaitoUTXOSetKey},
     time::create_timestamp,
     transaction::Transaction,
 };
 use serde::{Deserialize, Serialize};
+use ahash::AHashMap;
 
 extern crate rayon;
 use rayon::prelude::*;
@@ -102,6 +103,10 @@ impl Block {
         self.hash
     }
 
+    pub fn get_lc(&self) -> bool {
+        self.lc
+    }
+
     pub fn get_id(&self) -> u64 {
         self.core.id
     }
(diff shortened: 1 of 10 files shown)


escapedcat:
Regarding `fix: cargo fmt` commit message. There's a `style` type for this kind of commit because it only affects code-style.  
`fix` would bump a patch-update if we create releases automatically at one point in time. `style` won't do that.  

For these kind of commit just use i.e. `style: cargo fmt`

PR #136, opened 2021-06-29
Chore: cleanup for clippy warnings and mempool testing

BearGuy:
Just a little bit of cleanup from clippy warnings and trying to get some more tests in. Once we start solidifying other parts of the code, more tests will follow.


trevelyan (commented):
pub fn get_latest_block(&self) -> Option<&Block> {
        let block_hash = self.blockring.get_longest_chain_block_hash();
        self.blocks.get(&block_hash)

Some safety checks removed here -- what happens if the block doesn't exist?

PR #139, opened 2021-06-30
Feature: add slips to wallet and collect in add_block_sucess

BearGuy:
This still needs to be extended to `on_chain_reorganization`, but due to `wallet_lock` being behind an `async` `RwLock` makes the implementation complicated. Either `un/wind_chain` will have to be changed to manage `async` recursive functions, or the lock will have to be changed.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -1,12 +1,10 @@
 use crate::{
-    crypto::{
-        hash, MerkleTree, SaitoHash, SaitoPublicKey, SaitoSignature, SaitoUTXOSetKey, SHA256,
-    },
+    blockchain::UtxoSet,
+    crypto::{hash, MerkleTree, SaitoHash, SaitoPublicKey, SaitoSignature, SHA256},
     slip::SLIP_SIZE,
     time::create_timestamp,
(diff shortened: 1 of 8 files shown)


trevelyan:
Let's chat about this. There are a couple of things in here that we should go back to basics on.

1. The most important is that routing clients shouldn't process wallet payments by default. Looping through the block to check if we have received payments involves loops-within-loops. We can't afford to have our Rust client do this. Having a wallet with spendable slips is only necessary so we can generate transactions that spend fees and produce actual routing work.

2. Slips have on_chain_reorganization called as follows: blockchain.validate() -> block.validate() -> tx.validate() -> slip.validate(). They don't need that to happen in the wallet.

3. As part of this, I think we should make a clean division between what is in the wallet (references to spendable UTXO) and the concept of the slip as it exists in the blockchain. Without that distinction we get things like the wallet here calling on_chain_reorganization and asking for things like the UTXOset. It should not need that stuff.

4. I will add async to add_block in master now. That falls outside this wallet update and is important. good catch.

trevelyan:
Most of these changes have been merged into main by being ported into the merged sequential_refactor_wallet branch..

PR #140, opened 2021-06-30
Sequential refactor expanded class validation

trevelyan:
Block validation now takes place in two parts.

The first part requires writable blocks / txs and generates the data such as tx.hash_for_signature

The second part is read-only (validate).

Division is needed to provide the read-only process with access to the blockchain.


escapedcat:
@trevelyan I think in this case the Clippy error has a point, i.e.:
```bash
error: this comparison involving the minimum or maximum element for this type contains a case that is always true or always false
   --> src/transaction.rs:328:12
    |
328 |         if nolan_in < 0 {
    |            ^^^^^^^^^^^^
    |
    = note: `#[deny(clippy::absurd_extreme_comparisons)]` on by default
    = help: because `0` is the minimum value for this type, this comparison is always false
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#absurd_extreme_comparisons
```

trevelyan:
Yeah. Understand.

I didn't want to remove it because it's a necessary sanity check and want it to be there as a reminder we need to avoid this edge-case. Not sure if people can even provide negative values with Rust.

Should comment out though.

batzor:
@trevelyan It seems like it's not even an edge case since `nolan_in` is u64.

batzor:
Build error is also a must-fix. I suppose you forgot to delete the old implementation. I can understand that rustfmt check and convco is annoying sometimes but at least for the build check, I hope you could fix it before merging the PR :)

escapedcat:
@batzor agreed regarding build. `main` should always be able to build!  

We left this out on purpose before, because everything was kinda unstable some weeks ago. But by now I woudl rather turn this on to be required as well.

@trevelyan would this be ok with you?

PR #141, opened 2021-06-30
parallel merkle root generation

trevelyan:
This branch manually generates MerkleRoot instead of relying on an external library. Among the advantages:

* this uses the default Saito hashing algorithm instead of using Sha256
* rayon used for parallel generation of hashes, etc.

Timestamps below calculated on blocks with 10,000 txs each of which has a 1024 byte message field.

OLD
 ... start merkle: 1625035550570
 ... stop merkle:  1625035550714

NEW
 ... start merkle: 1625035717373
 ... stop merkle:  1625035717389

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -1,8 +1,7 @@
 use crate::{
     blockchain::Blockchain,
-    crypto::{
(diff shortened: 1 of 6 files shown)


clayrab:
LGTM

clayrab (approved):
LGTM

PR #144, opened 2021-07-01
chore: cleanup warnings

clayrab:
It's important that we keep the code warning-free. Warning are useful when developing because I use them to catch my own mistakes. This doesn't work well if there are other warnings already present.


trevelyan (approved):
thx

escapedcat:
At he moment Clippy is only failing for `errors`.  We can also make it fail for `warnings`.   
If there are cases where you know why you want to turn off a warning, you can turn of Clippy check individually on a specific function.

PR #147, opened 2021-07-01
feat: load transactions from disk

clayrab:
This is going to be difficult to work into add_block because it is &mut self.


clayrab:
The problem about &mut is a question of who owns the transactions after they are reloaded from disk. I think the most natural thing would be for the block to own them, which is what I've done. Then the interface is just get_transactions(), but this requires the block to be mutable, which it isn't in add_block. The other option would be for the block to keep track of it's state with some bool like "has_its_own_transactions" and if it's false then add_block goes to some other namespace that manages this, e.g. TransactionPool or something which just holds collections of tx keyed by block hash.

I will discuss more details of this with you today.

PR #152, opened 2021-07-03
Sequential refactor mempool tx pool

trevelyan:


trevelyan:
Mempool now generates blocks from a vector of transactions that is maintained in the mempool. When a block is created, that vector is memory-swapped with the vector in the block itself, speeding-up block creation.

the process of generating golden tickets and validating transactions and tickets has been streamlined.

PR #157, opened 2021-07-06
Feat/39 keypairstore

escapedcat:
Most simple draft I came up with.  

This will tell you how to set pw/path for keyfile:
```
cargo run -- --help
```

This will use defaults:
```
cargo run
```

This will use specfic file:
```
cargo run -- --key-path boom.txt
```


Cargo.toml
```diff
@@ -10,28 +10,30 @@ default-run = "saito_rust"
```
escapedcat, line 10: - adding rpassword and clap
- rest is just resorting A-Z

src/blockchain.rs
```diff
@@ -2,6 +2,7 @@ use crate::block::Block;
 use crate::blockring::BlockRing;
 use crate::consensus::SaitoMessage;
 use crate::crypto::{SaitoHash, SaitoUTXOSetKey};
+use crate::keypairstore::KeyPairStore;
```
escapedcat, line 5: Just adding this here because I wouldn't know how else to make it run automatically  ¯\_(ツ)_/¯

clayrab:
I've copied this code and re-implemented it as part of another PR. https://github.com/SaitoTech/saito-rust/pull/198.

PR #160, opened 2021-07-07
chore: update to latest rust version in GH action

escapedcat:
This will make sure we always use the latest version of Rust for the build-step in Github actions.  

Currently with this build works, but tests fail, like it does locally.

image


trevelyan:
I am ok with this if you guys are.

PR #162, opened 2021-07-08
chore: use gitignore instead of empty-file to keep dir

escapedcat:
It's nicer to handle an empty directory which should be kept using git instead of an empty file


trevelyan:
much better, thanks Hannes

PR #165, opened 2021-07-09
Review of wind unwind

trevelyan:
I believe this fixes the problems with wind/unwind chain. Would appreciate a review -- this and the add_block() function in blockchain are where we ran into the hardest problems in the javascript version. i hope that a lot of that was from the messiness of the code and that this implementation will be easier because it is cleaner.

In terms of the problems fixed, I think the conceptual messiness responsible for them originally was my use of the term current_wind_index/current_unwind_index in a way that suggested that the variable tracked BOTH [# of block in block_id sequence] as well as [index position in new_chain/old_chain vector].

I don't know if there is a better way to name those variables, but I have tried to comment-the-fuck out of the code so that it is always clear with specific examples that they should be pointing to the current position in the vector that we want to unwind / wind rather than something that must be [added_from_zero] or [substracted_from_vec_len] to find the correct hash we need to process now / next.

### DIFF
--- a/src/blockchain.rs
+++ b/src/blockchain.rs
@@ -387,7 +387,7 @@ impl Blockchain {
     //
     // when new_chain and old_chain are generated the block_hashes are added
     // to their vectors from tip-to-shared-ancestors. if the shared ancestors
-    // is at position [0] for instance, we may receive:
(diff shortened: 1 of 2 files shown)


trevelyan:
merging as necessary for next steps.

PR #167, opened 2021-07-10
Tests and cleanup

BearGuy:
Some additional tests and cleanup of functions without changing functionality too great.


trevelyan (approved):
I'm OK with this if it compiles against the sequential_refactor version that I'd like to merge into main that introduces slips, routing work and the ability to generate transactions that spend slips.

Much is an obvious improvement.

PR #172, opened 2021-07-13
Reviving spammer with transaction POST endpoint

BearGuy:
I've taken the `generate transactions` code in Mempool and revived the spammer to utilize it. Instead of dropping transactions straight into the `Mempool`, it POSTs them to an HTTP endpoint that creates the transaction and adds it into the `Mempool` over the broadcast channel.

Currently, there's one open question that I've uncovered while working on this, and that's around the "hash_for_signature" field in `Transaction`. Currently, it is not being serialized across the network, and I'm not sure if that's a design choice or not. It's required for generating the block's merkle root, so either.

1. It's de/serialized along with the other Transaction attributes
2. It's regenerated in deserialization
3. It's regenereted when fetched across the network

In cases of regeneration, I would advocate for the attribute to become an `Option` rather than have it default to an empty hash. In that situation, when the attribute is being fetched, it can be regenerated on the fly.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -31,6 +31,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Update Rust
+        run: rustup update
+      - name: Check Rust version
+        run: rustc --version
       - name: Build
         run: cargo build --verbose
       - name: Run tests
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -224,6 +224,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "245097e9a4535ee1e3e3931fcfcd55a796a44c643e8596ff6566d68f09b87bbc"
 
 [[package]]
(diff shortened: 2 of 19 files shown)


trevelyan:
> Currently, it is not being serialized across the network, and I'm not sure if that's a design choice or not. 

This is deliberate. The hash is generated from the message field, and confirming its validity is part of the block validation process. We can switch to an option if you'd like though.

The hash itself is regenerated during validation in the pre-validation-calculation function in which the transaction is writeable but also being processed in parallel. The hashing itself is computationally-intensive, so it should not be generated on-the-fly elsewhere unless we have an identified need for the data to exist. The mempool may need to run that function on transactions independently in order to generate / calculate the routing work in them.

BearGuy:
OK, I will be working on another PR that decides when `hash_to_signature` should be regenerated, alongside simplifying validation logic as a whole.

trevelyan (approved):
LGTM

PR #174, opened 2021-07-14
Initial Support : Automatic Transaction Rebroadcasting

trevelyan:
Includes some initial support for automatic transaction rebroadcasting.

We are not adding the ATR transactions to the block or validating them. But we are calculating them based on the assumption of a 2 block EPOCH. So this is scaffold code. What remains is basically to decide how to store the ATR transactions in the block so that we can have a fast and optimized sweep through the transaction set.

This PR also includes some updates to the spammer code that gets the transactions it is producing using Saito tokens owned by the wallet. And adds routing hops to them on receipt in the mempool so that we can calculate / test the routing work-related functions as needed.

I think this is ready for merge. The critical parts are largely contained to the Block.rs function that generates consensus data. And it fixes a number of bugs introduced by our merging of previous pull requests from different branches yesterday.

### DIFF
--- a/src/bin/spammer.rs
+++ b/src/bin/spammer.rs
@@ -60,15 +60,13 @@ pub async fn main() -> saito_rust::Result<()> {
                 // sign ...
                 transaction.sign(privatekey);
 
+		// make sure we know fees etc.
+		transaction.pre_validation_calculations_parallelizable(publickey);
                 transactions.push(transaction);
 
 	    }
 
             for mut tx in transactions {
-
-	        tx.add_hop_to_path(wallet_lock.clone(), publickey).await;
-	        tx.add_hop_to_path(wallet_lock.clone(), publickey).await;
-
                 let bytes: Vec<u8> = tx.serialize_for_net();
                 let _res = client
                     .post("http://localhost:3030/transactions")
--- a/src/bin/spammer.rs
+++ b/src/bin/spammer.rs
@@ -1,9 +1,7 @@
 use saito_rust::{
     blockchain::Blockchain,
-    crypto::{generate_random_bytes, hash, SaitoPrivateKey, SaitoPublicKey},
     mempool::Mempool,
     miner::Miner,
-    slip::Slip,
     transaction::Transaction,
     wallet::Wallet,
 };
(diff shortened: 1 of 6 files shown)


src/bin/spammer.rs
```diff
@@ -46,9 +44,29 @@ pub async fn main() -> saito_rust::Result<()> {
         let client = reqwest::Client::new();
         // sleep(Duration::from_millis(5000));
         loop {
-            let transactions =
-                generate_transactions(publickey, privatekey, txs_to_generate, bytes_per_tx);
-            for tx in transactions.iter() {
+
+            let mut transactions: Vec<Transaction> = vec![];
+
+	    for _i in 0..txs_to_generate {
```
BearGuy, line 50: Overall, PR is looking good. One thing I'll request before merging this is to fix the formatting. You can see in files changed how poor the formatting can get in a bunch of places. I'm sure it looks fine in your dev environment, but for everyone else, the code becomes messy and strange. I'd like to request that you fix the spacing in your environment so that we can have proper spacing in the repo.

BearGuy (changes requested):

src/block.rs
```diff
@@ -503,101 +521,115 @@ impl Block {
...
-            //
-            // calculate miner and router payments
-            //
+		//
+                // winning router is picked by sending a random 
+		// number into the transaction, which is then 
+		// used to select a routing node based on the 
+		// weighted lottery.
+                //
+            	let random_number2 = hash(&miner_random.to_vec());
```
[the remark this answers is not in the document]
trevelyan: The first random number is the miner_random.

Logic here is going to get more complicated as we add in staking, so lets come back and fix variable names when we add that!

src/bin/spammer.rs
trevelyan: I've noticed these issues. The problems were especially severe in the mempool.run() function and the GenerateTransaction logic that has now been moved into that spammer class. I suspect the "cargo fmt" formatter was struggling with the tokio/async loop internals, as well as cases within a function where we have code that looks like this -- with read / write access compartmentalized so that the lock releases quickly:

   let publickey;

  {
      let wallet = wallet_lock.write().await;
      publickey = wallet.get_publickey();
 }

The problems may also be getting caused by the Github web-interface for sorting out merges, since when the files load in that web interface they are often mangled in terms of how they display and how the different sections align. I've been using that to handle those fixes and can switch to another approach if there is something better.

At this point I've cleaned up all of those files several times. It's a bit of a disinhibitor to run "cargo fmt" when you know you might have to re-indent a lengthy function. I'm happy for us to run cargo fmt by default on code. What I try to do is clean up code when I'm looking at it in VIM and it is badly aligned. If the issue is others having a different standard for how many spaces are in a tab (I believe VIM is 8) I would be happy to switch over.

Otherwise, the best way to avoid the problem would seem to be not having large changes that sweep over multiple classes and require manual merge.

PR #189, opened 2021-08-03
Command Line Interface

clayrab:
I've done a quick CLI which is very helpful when trying to test network interfaces. I've used a library called Clap which Hannes discovered and it looks quite professional so we can definitely extend this in the future for users.

See the markdown for details how to use it.


clayrab:
I

PR #197, opened 2021-08-09
Basic Config File for Network Settings

BearGuy:
Basic config being set for a network's host and port


src/networking/network.rs
```diff
@@ -374,4 +398,170 @@ mod tests {
         let mempool = mempool_lock.read().await;
         assert_eq!(mempool.transactions.len(), 1);
     }
+
+    fn parse_response(message: Message) -> (String, u32, Vec<u8>) {
```
clayrab, line 402: We already have the APIMessage struct and APIMessage::deserialize

PR #209, opened 2021-08-25
fix: remove broken filename memoization and force blocks to be written to disk

clayrab:


clayrab:
I'm going to merge this and open a bug. There is a problem with this code that hasn't actually been fully resolved, the block is being forcibly saved, but we need a more flexible solution. I believe it should be flexible enough to deal with writing to disk during testing and should also address the "blocks are being saved twice" issue which Stephen flagged last week.

PR #218, opened 2021-09-01
Fix: save_to_block flag removed and replaced with path existence check

BearGuy:
After today's meeting, I removed the additional argument in the `add_block` for determining whether we try to save the block to disk. It has been replaced instead with a check to see if the file already exists, and if it does it will not attempt to write to disk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ This project will serve as the reference implementation for other code-bases. It
 
 ## Contributing
 
-We're happy for any contribution.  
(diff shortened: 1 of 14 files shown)


clayrab:
Sorry, we took too long to merge this and it became very difficult to merge.

David merged this by cherry picking out the relevant bits and merging it into main in another PR.

We also ran cargo fmt and cleaned up the warnings to main is in a place where this should be less of a problem in the future.

PR #232, opened 2021-09-13
feat: the cli tool now has  ability to read block data from disk

clayrab:


src/bin/saitocli.rs
```diff
@@ -40,27 +40,79 @@ or
 ```bash
 cargo run --bin saitocli -- tx -a 1 -t gYsu1fVHjP6Z8CHCzti9K9xb5JPqpEL7zi7arvLiVANm -f tx.out
 ```
+or
```
escapedcat, line 43: should this be moved to the readme?

PR #240, opened 2021-09-14
test: added some tests to mempool

clayrab:


src/mempool.rs
```diff
@@ -442,15 +431,14 @@ mod tests {
         assert_eq!(Some(block), mempool.blocks_queue.pop_front())
     }
 
-    #[ignore]
     #[tokio::test]
-    async fn blockchain_test() {
+    async fn mempool_bundle_blocks_test() {
         // TODO Fix this test
```
escapedcat, line 436: is this fixed now?

clayrab: ah, yes, thx. good catch.

PR #243, opened 2021-09-16
Block serialization tests

clayrab:


escapedcat:
@clayrab this will fial the commit linting:  
image  
Please only use one type, i.e. `fix` or `refactor`. Both is not possible following https://www.conventionalcommits.org

PR #248, opened 2021-09-20
fix: tests for mempool block production

trevelyan:
quick addition of mempool test requested by Clay.

pushing quickly into main as was in this branch when I made hte fix.


escapedcat:
@trevelyan would be great if you could run `cargo fmt` before commiting. It's easy and helps to keep code-style aligned and the check green :)

trevelyan:
thx, will try

Issue #259, opened 2021-09-28
Suggest using the updated jemalloc lib


thaodt:
Hi team, I've just noticed that we're using the outdated [jemallocator](https://github.com/gnzlbg/jemallocator).
Check [this PR](https://github.com/rust-lang/rust/pull/83152) from rust-core for more details.

**Updated jemalloc lib**
This updated one is from [tikv](https://github.com/tikv/jemallocator)

TL;DR: this upgrade could help for sized deallocation's usage correctly, particularly to rustdoc.
So somehow it can deliver a sizable performance increase. We can do some benchmarks after that.

Lets consider my suggestion. Thank you.

clayrab:
Thanks so much for pointing this out, Thao.

The reason I added jemalloc in the first place was related to jemalloc-ctl because I wanted to do some introspection on memory usage in the old codebase... we aren't currently using this because the debate about memory usage has basically been settled...

If I'm understanding the discussions linked above correctly, it seems that rust-lang is already using tikv-jemallocator, so perhaps we should just remove both jemalloc packages from the toml and delete memory-stats.rs?

That being said, if this is mostly an improvement for rustdoc, I'm not particularly concerned about that unless it would improve compile times during dev... but in theory that shouldn't be the case, right?(i.e. why would be build the docs on every compile?).

I think the proper way to do this would be as Thao suggested and do some very basic benchmarks, at least. However, I wouldn't be opposed to simply removing the packages, since that's probably the right thing to be doing anyway. If no one else wants to handle this, I'll probably just proceed this way.

thaodt:
> Thanks so much for pointing this out, Thao.
> 
> The reason I added jemalloc in the first place was related to jemalloc-ctl because I wanted to do some introspection on memory usage in the old codebase... we aren't currently using this because the debate about memory usage has basically been settled...
> 
> If I'm understanding the discussions linked above correctly, it seems that rust-lang is already using tikv-jemallocator, so perhaps we should just remove both jemalloc packages from the toml and delete memory-stats.rs?
> 
> That being said, if this is mostly an improvement for rustdoc, I'm not particularly concerned about that unless it would improve compile times during dev... but in theory that shouldn't be the case, right?(i.e. why would be build the docs on every compile?).
> 
[...]

PR #264, opened 2021-09-30
fix: remove jemallocator package in favor of rust-lang default.

clayrab:
We were using package jemallocator for memory-stats.rs. Letting rust use it's default is supposedly more performant according to @thaodt and it's likely safer and "the right thing to do", as well. In lieau of doing a bunch of benchmarking and overthinking this, let's just remove it right now and we can revisit if/when we want to re-do benchmarking work.


escapedcat:
#259

PR #270, opened 2021-10-01
Unbottlenecking the Network

trevelyan:
CRITICAL:

* "unbottlenecks" progress on networking code by eliminating issues flagged preventing multi-node sync

ALSO 

* network run() invocation and methods restored to cross-class standards
* eliminates /bin/spammer.rs, moving tx-generation into the main codebase
* pre-emptive block.generate_hashes() in blockchain.add_block(), to ensure hash consistency at bottleneck where it matters
* blocks successfully loading from disk
* disables wallet keyfile and password loading

REQUIRES:

 * review and improvement of CLI argument handling so that the wallet may OPTIONALLY be provided and tx-generation can be made an optional feature.


escapedcat:
can haz rustfmt?

src/consensus.rs
```diff
@@ -224,10 +256,9 @@ impl Consensus {
         // Network
         //
             res = crate::networking::network::run(
-                settings,
```
escapedcat, line 227: `settings` not needed anymore?

trevelyan:
it's sent over when the network object is initialized, so we don't need to
include it again as a function argument.

just keeps things a bit simpler.

trevelyan:
Hi Markus,

Yeah, please. Go for it!

--david

PR #272, opened 2021-10-04
docs: remove outdated spammer paragraphs

HerrBertling:
As commented in #270 

I figured opening the PR would be faster than waiting for anyone to answer my question, so here you go 😄


trevelyan (approved):
valid

PR #274, opened 2021-10-05
Wallet file and settings management

trevelyan:
This push updates the wallet class to add some changes:

* STORAGE *

Wallet files are now stored in ./data/wallets

* INITIALIZATION *

Running the main codebase will always initialize a wallet. This eliminates problems with the password prompt on startup making the main interface less usable for testing and development purposes. It also avoids the need to have an encrypted test wallet or distribute the code with one in order for users / devs to be able to .

The arguments -wallet [filename] and --password [password] can be used to feed information to the binary and tell it which wallet to use. If we want to use STDIN prompts to fetch the password as an option I would suggest we add a command-line option that permits this.

If Saito is started without a wallet explicitly provided the software will create a temporary one on start that will be saved as ./data/wallets/default. This wallet will be over-written every time the software is restarted in this fashion, but the default wallet created is not deleted on close so that keys may be recovered if necessary. All wallets on disk are encrypted with a password -- the default password for all wallets where a password has not been provided is "password".

If the name of a wallet is provided the wallet will be loaded from disk if a wallet with that name exists in the ./data/wallets directory. If a wallet with that name does NOT exist the default wallet that is being created will be saved as that file. The logic for creating and saving these wallets has also been simplified so that wallet objects can simply call save() and load() and data updating will be handled automatically.

* CRYPTO *

The previous wallet file had in-class code that handled AES symmetrical encryption of the privatekey prior to its being saved to disk. The code worked perfectly and is what we need. As crypto functions belong in the crypto.rs class I have moved that functionality over to crypto.rs where symmetrical encryption is now provided as a generic service on bytes-vectors. The two functions encrypt_with_password and decrypt_with_password are what we should use, although if anyone wants to suggest a better name that is fine. In combination with save() and load() this shift has allowed us to simplify wallet creation considerably.

* SERIALIZATION / DESERIALIZATION *

Wallet files are now serialized and deserialized through dedicated functions in the Wallet class that mirror the way other components are serialized / deserialized. This has permitted removal of the Persistable trait and some Rust-specific code and as such I believe provides for greater wallet portability. The main reason I think this is a good idea is that it eliminates the STRUCT EncryptedWallet completely, and permits the wallet to save additional kinds of data, such as spendable slips. Unless we are planning to have the config file writable, it probably makes sense to have the wallet store information such as the latest_block_hash and latest_block_id the software has seen, as well as persisting the fork_id over starts.


escapedcat:
> The arguments -wallet [filename] and --password [password] can be used to feed information to the binary and tell it which wallet to use. If we want to use STDIN prompts to fetch the password as an option I would suggest we add a command-line option that permits this. It's obviously useful for cases where users do not want to hardcode the password onto startup files, but doesn't need to be the default behavior. I've removed it for now because it isn't critical to getting Saito working properly.

> If Saito is started without a wallet explicitly provided the software will create a temporary one on start that will be saved as ./data/wallets/default. This wallet will be over-written every time the software is restarted in this fashion, but the default wallet created is not deleted on close so that keys may be recovered if necessary. All wallets on disk are encrypted with a password -- the default password for all wallets where a password has not been provided is "password".

> If the name of a wallet is provided the wallet will be loaded from disk if a wallet with that name exists in the ./data/wallets directory. If a wallet with that name does NOT exist the default wallet that is being created will be saved as that file. The logic for creating and saving these wallets has also been simplified so that wallet objects can simply call save() and load() and data updating will be handled automatically.

Wanna add this to the README?

PR #279, opened 2021-10-13
fix: saitocli wallet stopped working after chagnes to wallet.

clayrab:


escapedcat:
Typo in commit message ;)

PR #284, opened 2021-10-25
devops: add log and replace println.

clayrab:


clayrab:
This snuck in with another PR... oops

PR #294, opened 2021-11-02
feat: blockchain signal network to propagate blocks.

clayrab:
To try this, from one repo do: cargo run -- --spammer

This should give you blocks and Warp should start listening for socket connections.

if you want to start over, do: rm data/blocks/*.sai

From another repo, do: cargo run -- --config=config2.yml --password=asdf

This will try to connect to the peer specified in config2.yml(which is the default port which we ran above with cargo run).

The 2nd node should connect and sync itself with the first node, then when the first node makes a block, we want the NetworkNewBlock message to trigger a SNDBLKHD API message and to 2nd node which should then do a REQBLOCK back to the first node.

### DIFF
--- a/src/blockchain.rs
+++ b/src/blockchain.rs
@@ -348,7 +348,7 @@ impl Blockchain {
             self.broadcast_channel_sender
                 .as_ref()
                 .unwrap()
-                .send(SaitoMessage::NetworkNewBlock { hash: [0; 32] })
+                .send(SaitoMessage::NetworkNewBlock { hash: block_hash })
                 .expect("error: BlockchainAddBlockFailure message failed to send");
         }
         trace!(" ... block save done:            {:?}", create_timestamp());
--- a/src/blockchain.rs
+++ b/src/blockchain.rs
@@ -344,7 +344,13 @@ impl Blockchain {
                 Storage::write_block_to_disk(block);
             }
         }
-
+        if self.broadcast_channel_sender.is_some() {
+            self.broadcast_channel_sender
+                .as_ref()
+                .unwrap()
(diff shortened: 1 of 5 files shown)


escapedcat:
Wanna add this to the readme?

trevelyan (commented):
small feedback -- the convention for SaitoMessage is that the name is the class where the event is fired, and then the activity or event that is triggering the broadcast. This is supposed to make it easy to figure out at a glance what has fired the event when looking at code that is listening in a separate class.

See:

    BlockchainNewLongestChainBlock { hash: SaitoHash, difficulty: u64 },
    BlockchainAddBlockSuccess { hash: SaitoHash },
    BlockchainAddBlockFailure { hash: SaitoHash },
    MinerNewGoldenTicket { ticket: GoldenTicket },
    MempoolNewBlock { hash: SaitoHash },
    NetworkNewTransaction { transaction: Transaction },

This approach also lets us piggyback multiple actions in different classes onto a single event. Both module and miner components can listen for BlockchainNewLongestChainBlock, for instance, instead of the Modules class having to listen to a MinerMineNewHash message that is fired by the Blockchain and suddenly devs are grepping to figure out what is firing these messages and when and where, etc. Cleaner and more intuitive.

So I would suggest updating to BlockchainBlockSaved or something like that.

clayrab:
Sanka is maybe going to take this PR the next mile. Not sure if we will use this branch/PR or he'll start over. We're just getting him setup on github.

batzor:
Will it spam transactions?

clayrab:
> Will it spam transactions?

We are sorting this out this sprint. We'll basically be spamming block headers to people who may alreayd have the block for now. It's a decent first solution.

thaodt:
LGTM.

PR #296, opened 2021-11-03
Feature/sanka/printing logs to std streams

SankaD:
Now developer can choose which logs to be shown on console when running unit tests.


clayrab (commented):
LGTM

clayrab (approved):

PR #297, opened 2021-11-04
feat: init tests suite

thaodt:
I jump into this idea because we’re going to build all our test cases into a single crate.
By default, rust will compile each of `#[test]` labeled functions into its own binary crates (with its own main()) then executes it as part of the test.
This one is just a first step to remove `--test-threads=1` from `cargo test`.
I will add some configuration scripts for this. 
In the meantime, lets see what I have in it.
How to use: `cargo test ts`

### DIFF
--- a/ts/main.rs
+++ b/ts/main.rs
@@ -1,0 +1,22 @@
+pub mod tests;
+use tests::TestsSuite;
+
+fn setup() {
+    println!("----------------- Setup -----------------")
+}
+
+fn teardown() {
+    println!("----------------- Teardown -----------------")
+}
+fn main() {
+    // Setup test env
+    setup();
+
+    // Run test cases
+    for t in inventory::iter::<TestsSuite> {
+        (t.test_sth)()
+    }
+
+    // Teardown test env
+    teardown();
+}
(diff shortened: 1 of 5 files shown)


ts/main.rs
```diff
@@ -0,0 +1,20 @@
+pub mod tests;
+use tests::TestsSuite;
+
+fn setup() {
+   println!("----------------- Setup -----------------")
```
escapedcat, line 5: In this case it ok to use `println` and not the logger? Not pointing out, just asking to learn how this is done in rust.

thaodt: image
No, in the setup & teardown functions, we can put sth like setup logger served for testing. Loading test data at `setup()` step, and do clean up on `teardown()`. Its more like my demo about tests suite.

clayrab:
let's delete this "ERROR: old chain length is greater than new chain length" message. Also, if we could get this into a git action, that would be awesome.

When I run this, it seems to hang on the test "test_manager_blockchain_fork_test"...

clayrab:
Also, we can get rid of the test_be_sure_threads_equals_1 test if this works...

escapedcat:
Adjusting the README would be nice :)

thaodt:
Close this PR. I made another one.

PR #298, opened 2021-11-04
fix: Other removed from TryFromByte enums

clayrab:
These were needed when we only had Normal and Other types because TryFromByte required at least 2 fields in the enum. They are no longer needed.


thaodt:
Lgtm.

trevelyan:
agree

trevelyan (approved):
yes

PR #299, opened 2021-11-08
fix: golden ticket algorithm updated for 1/16 reduction per difficult…

trevelyan:
This push fixes the Golden Ticket Algorithm. The previous one was not working in the sense that any solution would be considered valid, so it was fine for testing, but the issues would have become apparent running longer-chains on a working network.

The code itself is not brilliant. It uses String to recreate a SaitoHash in HEX and then converts it to  binary in order to do the difficulty comparison. I'm sure the push can be cleaned up and simplified, but am leaving it for now as the code works. It should be possible to make the code more compact, and I would welcome tests being written against it, but am submitting as-is in order to be able to move-on with the Saito-Lite-Rust development.

In terms of the substance of difficulty calculation, what we are doing is generating a target HEX at the maximum acceptable difficulty and then comparing the randomly-calculated difficulties against is. On a 32-byte hash this means 64-HEX digits, each representing 4-bits. The algorithm starts with a difficulty of 1, which is:

 EFFFFF...

As difficulty rises, it drops:

 DFFFFF...
 CFFFFF...
 BFFFFF...
 
 We do this instead of working with bit-level 0s and 1s in order to keep the algorithm the same as the Javascript implementation for easy comparison of correctness. We can always shift to bit-level flicking of 1 --> 0 if people want to eliminate the String manipulation completely.
 
One final note -- this push also updates the inputs that are provided when generating the Golden Ticket -- they must include the previous_block_hash AND a random nonce AND the publickey of the node that will be making the claim on the mining solution.

Best,

--david


clayrab:
I'm afraid this will comprimise our next milestone since some of the fork-related tests are failing. recommend that we do not merge this until that's resolved, but I'm not sure if this is necessary for the next milestone or not

trevelyan:
The above pushes fix the test failures. Explanation of the problem briefly below, in case anyone is curious:

Previously all submitted solutions were being deemed valid.

With the new algorithm, an increase in the difficulty made it 1/16 harder to produce a valid solution. 

The function miner.mine_on_block_until_golden_ticket_found() is used in our test utilities to hash a golden ticket solution until one exists. This function is used for testing chain-reorganizations involving payouts, as we sometimes need to specify that a payment should happen every N blocks.  If we use "generate_blockchain" or "generate_block" in the TestManager we are calling this function when we insist that a block includes a golden ticket.

The golden ticket was being created properly, but then a different hash (the current target on the main chain) was being swapped into the solution instead of the target_block_hash (which could be several blocks deep on a different chain). The result was that we ended up with a different "solution"., And whether that solution would pass was basically a question of whether a random new hash would pass.

Previously the algorithm let everything pass, so we did not run into this problem. With the revision, a certain percentage of new blocks would fail -- and a greater and greater percentage depending on how high the difficulty ran. So we started running into test failures on long-chain tests that would produce enough golden tickets to raise difficulty. But the failures would be unpredictable because they would only start by affecting 1/16 of blocks produced. So things would not regularly fail but then suddenly fail.

I'm not sure whether there is a good way to test for this sort of stochastic (?) failure. But the updates fix the problem by ensuring that the golden ticket included in the block is the same one that was generated and confirmed to pass difficulty criteria.

clayrab (approved):
Looks okay. I think we need to get some kind of "goldens" solution working if we really want to deal with hashing properly.

PR #302, opened 2021-11-09
bugfix: loading blocks of another chain from disk

SankaD:
added some tests to find the bug


clayrab (approved):
Might as well merge this. We can add more assertions or coverage later ,but there's no reason not to add this.

PR #303, opened 2021-11-09
feat: graceful shutdown warp

thaodt:
The warp port does not shutdown properly when we issue interrupt signal to the binary. Therefore, I used [this warp function](https://docs.rs/warp/0.3.1/warp/struct.Server.html#method.bind_with_graceful_shutdown) for graceful shutdown feature.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /src/data
 *.swp
 .idea/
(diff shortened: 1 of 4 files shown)


clayrab (approved):
lgtm

PR #315, opened 2021-11-26
feat: tx propagation

thaodt:
Transaction Propagation - SNDTRANS. Im writing tests for this. But you can help me review the main logic. Thank you.


src/networking/peer.rs
```diff
@@ -480,8 +480,13 @@ impl SaitoPeer {
             }
             "SNDTRANS" => {
                 if let Some(tx) = socket_receive_transaction(api_message.clone()) {
+                    let blockchain = blockchain_lock.read().await;
                     let mut mempool = mempool_lock.write().await;
-                    mempool.add_transaction(tx).await;
+                    if mempool.transaction_exists(tx.get_hash_for_signature()) != true {
```
clayrab, line 485: I'd rather just do !mempool.transaction_exists(tx.get_hash_for_signature()) than != true...

src/networking/peer.rs
```diff
@@ -480,8 +480,13 @@ impl SaitoPeer {
             }
             "SNDTRANS" => {
```
clayrab, line 481: I think there is some missing behaviour here. When we receive the transaction, we need to do something like propagate_transaction_to_peers, except there is some extra logic, for each peer we need to check if they are already in the path, if so, then we do not propagate to that peer.

src/networking/network.rs
```diff
@@ -346,6 +364,10 @@ impl Network {
                         warn!("SaitoMessage::BlockchainSavedBlock recv'ed by network");
                         Network::send_my_block_to_peers(block_hash).await;
                     }
+                    SaitoMessage::NetworkNewTransaction { transaction: tx } => {
```
clayrab, line 367: I think the intention which David has was for these message to indicate where they are coming from, so "NetworkNewTransaction" would imply that the message is coming from the SNDTRANS handler. We also need a way to handle transactions which we've created "ourself", which I'm now realizing is not exactly clear with saitocli... I guess this would really be from the wallet, e.g. SaitoMessage::WalletNewTranasction, so perhaps for now when we create tx via saitocli they should go through the wallet(even though we won't actually be using the wallet yet to get the inputs, we will in the future do something like this)

src/networking/network.rs
```diff
@@ -299,6 +300,23 @@ impl Network {
...
     }
+
+    // For transaction propagation made by mempool to all peers
+    async fn propagate_transaction_to_peers(tx: Transaction) {
+        let peers_db_global = PEERS_DB_GLOBAL.clone();
+        let mut peers_db_mut = peers_db_global.write().await;
+        // We need a stream iterator for async(to await send_command_fire_and_forget)
+        let mut peers_iterator_stream = futures::stream::iter(peers_db_mut.values_mut());
+        while let Some(peer) = peers_iterator_stream.next().await {
+            if peer.get_has_completed_handshake() && !peer.is_in_path(tx.clone()) {
```
clayrab, line 311: Let's try to get rid of the clone() here... It shouldn't be needed, we are only reading the tx, right?

src/networking/network.rs
```diff
@@ -299,6 +300,37 @@ impl Network {
...
+        wallet_lock: Arc<RwLock<Wallet>>,
+        mut tx: Transaction,
+    ) {
+        tokio::spawn(async move {
+            let peers_db_global = PEERS_DB_GLOBAL.clone();
+            let mut peers_db_mut = peers_db_global.write().await;
+            // We need a stream iterator for async(to await send_command_fire_and_forget)
+            let mut peers_iterator_stream = futures::stream::iter(peers_db_mut.values_mut());
+            while let Some(peer) = peers_iterator_stream.next().await {
+                let mut _path = tx.get_path().clone();
```
clayrab, line 315: Can we just skip the clone and pass &tx.get_path() directly to is_in_path?

clayrab: Also, why is this 'mut' anyway?? shouldn't the compiler be complaining about that? anyway, I don't think ti's needed at all.

clayrab (approved):
LGTM

PR #323, opened 2021-12-19
fix: no panic on verify()

batzor:
Now, the block verification will cause panic when the block creator field is ill-formed. Such block can cause all recipient clients to panic so I believe it is a critical fix.

- safer handling of ill-formed fields on verification.


thaodt:
Thank for pointing it out @batzor. LGTM.

Issue #328, opened 2022-01-03
Update and match the Onchain Reorganization behavior in Saito-Rust with Saito-lite-rust


karlsaito:
Acceptance Criteria:

- [x] Onchain reorganization in Rust is identical with Saito-lite-rust implementation
- [x] Passed the Tests after changes.

Subtasks
- [x] move out logic relating to `update_genesis_period()` & `fork_id` to blockchain `on_chain_reorganization()` itself.
- [x] `add_block_success()` update.
- [x] passed the tests.

PR #334, opened 2022-01-07
Feature: on chain reorganization update

thaodt:
Update and match the Onchain Reorganization behavior in Saito-Rust with Saito-lite-rust #328

### DIFF
--- a/src/blockchain.rs
+++ b/src/blockchain.rs
@@ -371,7 +371,6 @@ impl Blockchain {
             " ... blockchain.add_block_success: {:?}",
             create_timestamp()
         );
-        let block_id;
         // print blockring longest_chain_block_hash infor
         self.print();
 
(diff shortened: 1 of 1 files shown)


src/blockchain.rs
```diff
@@ -371,18 +371,32 @@ impl Blockchain {
...
         {
             let block = self.get_mut_block(&block_hash).await;
-            block_id = block.get_id();
             if block.get_block_type() != BlockType::Header {
                 Storage::write_block_to_disk(block);
             }
         }
+
+        //
+        // TODO: clean up mempool - I think we shouldn't cleanup mempool here.
```
escapedcat, line 388: To have an overview about open TODOs we tend to create tickets and link those in the code. Just an idea. /cc @karlsaito

trevelyan (approved):
LGTM

